PR #890, opened 2014-03-31
Fix page

passy:
Damn, this was really broken. I'm glad I noticed this.

### DIFF
--- a/site-assets/main.js
+++ b/site-assets/main.js
@@ -2,6 +2,13 @@
 (function () {
 	'use strict';
 
+	var date = new Date();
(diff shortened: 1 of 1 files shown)


site-assets/main.js
```diff
@@ -2,6 +2,13 @@
 (function () {
 	'use strict';
 
+	var date = new Date();
```
passy: 2014 passy really didn't give a flick about side effects ...

blakechambers:
This still seems to be fixed. 👏 👏 👏

PR #964, opened 2014-08-08
Dragome TodoMVC initial commit

fpetrola:
This an implementation of TodoMVC application for Dragome SDK (www.dragome.com).

Dragome is an open source tool for creating web applications in pure Java language. Based on bytecode to javascript compilation, you can execute applications written in Java directly on browsers. 
Using the template engine combined with two way binding mechanism we can produce a very short implementation of the application.
There is just one small class for building the entire UI, and two other classes that represent Todo Item model and Todo Manager model.
HTML template is almost the same template that TodoMVC specification provides, the only difference are "data-template" attributes added to 16 tags to identify each dynamic part.
UI is created using component builders, which allows us to create the component, associate it to a template part, add event listeners, and repeat subtemplates if required.
Model binding is managed by a combination of Form Bindings module (based on gwt-pectin) and Method Logger (bytecode instrumentation plugin) for detecting getters and setters access. 
It's not required the use of any explicit mechanism to detect changes over the model or UI, the built in two ways data binding takes track of each expression and execute it again when related value changes.


samccone:
hi @fpetrola 

if you do the following

```
git remote add real git@github.com:tastejs/todomvc.git
git fetch
git reset --hard real/master
git cherry-pick  feecbaf
```

and then `git push -f` your branch will only have one commit

Once that is done we can take a look and run the test suite, thanks again :santa:

samccone:
hi @fpetrola 

see my comment on https://github.com/tastejs/todomvc/pull/964#issuecomment-78097064 for how to fix up this branch. simply reopen the PR when you are ready.

thanks a bunch!

PR #1076, opened 2014-11-25
lazojs example

jstrimpel:
This is a submission for a [LazoJS](https://github.com/walmartlabs/lazojs) example.

***Marketing Blah:*** LazoJS is a client-server web framework built on Node.js that allows front-end developers to easily create a 100% SEO compliant, component MVC structured web application with an optimized first page load.

Essentially it is a fully baked client-server JS framework with sensible defaults for creating public facing sites. It transitions to an SPA after the initial page load. Not much focus was placed on the finer grained UI pieces. These details are left up to the application developer. Although, there is [support](https://github.com/walmartlabs/lazojs/wiki/Imports) for HTML imports, which can be used to leverage custom elements. The primary focus is providing the structure and foundation for creating large applications across multiple teams.

Please refer to the wiki for a high level [overview](https://github.com/walmartlabs/lazojs/wiki/Overview) and further information.


samccone:
Hi @jstrimpel thanks for the PR and the work in putting this together.
the issue of client side apps rendered on the server (for SEO) is a really difficult a tricky issue. Due to the complexity around this issue we have seen tons of plugin and standalone solutions pop up over the past few years, from Rendr to server side React. 

LazoJS seems like a really neat solution for a server side rendered SPA approach that works out of the box. That said, since it requires a server to work at all, it goes against the static only approach that the main TODO site has been following. 

If you want to stand up a server to host it, we will gladly link to it from our wiki so that people can still find and use this example. 

Thanks again and great work!!

PR #1082, opened 2014-12-05
TypeScript - KnockoutJS example

jj09:
I created an Todo list with TypeScript and KnockoutJS. As a part of this contribution I did small fix in existing KnockoutJS Todo.


samccone:
@jj09 Great points about how the combination of TypeScript and Knockout can lead to several gotchas. 

While this is a useful thing for developers looking to go down this path, I do not think it justifies its own spot as a unique example on the site, instead we would be happy to add it to the wiki page, 

Thanks for the work! :heartbeat:

PR #1087, opened 2014-12-12
fixing memory leak in dojo-based Todo

masrud:
This PR fixes a memory leak in Todo applicaiton developed using Dojo.


masrud:
Hi there,
I was wondering what was your thought on the fix proposed in this  pull request?
I understand the todo application has changed significantly since then, I mean the fix for that specific release.

Thanks,
Masrud

PR #1093, opened 2014-12-16
Added Backbone Dispatcher example

talyssonoc:


arthurvr:
In a duplicate PR, @passy [commented](https://github.com/tastejs/todomvc/pull/1094):

> To apply the same scrutiny to this PR as to all others, there are some remaining issues. The main backbone-dispatcher library is still very young and while based on an accepted pattern not battle-tested on its own. The lack of tests on the library itself is a bit worrisome. I'd prefer us to wait a little longer and see how the community reaction and pick up rate of it is before committing to including it.

I am totally +1 on that. [The project didn't reach any release yet,](https://github.com/talyssonoc/backbone-dispatcher/tags) so I would be :+1: on waiting till it matured a little, and then we could take a second look. @tastejs/todomvc What does the rest think?

addyosmani:
> I am totally +1 on that. The project didn't reach any release yet, so I would be :+1: on waiting till it matured a little, and then we could take a second look

Agreed

PR #1097, opened 2014-12-30
Add Jet-Angular Realtime example

lipp:
New realtime implementation. Please visit http://www.jetbus.io/ for more information about Jet.
The Jet [todo-server](https://github.com/lipp/node-jet/blob/master/examples/todo-server.js) is hosted at nodejtsu.


arthurvr:
Hi @lipp! Thanks for working on this. At the moment we feel like we're not going to include this app, sorry. Once the framework has matured and reached its first release, we can always reconsider this.

PR #1107, opened 2015-01-24
Updating Meteor to the newest version

michalsnik:
Slight change in Meteor example, that updates Meteor to the newest release and fix handlebars in templates  to work properly with it.
This change makes it possible to add new items in todo list one by one, without need to focus input manually after each insert. 
Now application can behave like any other example - clearly identical.


sindresorhus:
Thanks for the update :)

PR #1127, opened 2015-02-05
Fixes #1120 - Update of tpl.css to match current style

### LINKED ISSUE #1120: `Aria templates` example overwriting styles
Followup on https://github.com/tastejs/todomvc/pull/1116. Seems like the Aria templates example is overwriting/forcing [some of the old UI](https://github.com/tastejs/todomvc/blob/master/examples/ariatemplates/js/view/TodoStyle.tpl.css#L47).

olaf-k:
Long story short, AT has a mechanism to scope CSS rules that required me to rewrite most of the project css in a tpl.css.  This file has been updated to reflect the latest style changes.

### DIFF
--- a/examples/ariatemplates/js/view/TodoStyle.tpl.css
+++ b/examples/ariatemplates/js/view/TodoStyle.tpl.css
@@ -5,6 +5,31 @@
 
 {macro main ()}
 
+#label-toggle-all {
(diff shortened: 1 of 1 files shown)


passy:
Thanks for the update, @olaf-k!

PR #1129, opened 2015-02-06
add todomvc demo use avalon js && oniui

gogoyqj:
I have add a todomvc demo use [avalon js](http://avalonjs.github.io/#) and [oniui](http://ued.qunar.com/oniui/home.htm)


passy:
@gogoyqj Are there docs available in English? We want this to be as approachable as possible and without those, the majority of our users won't have the ability to compare this to other examples.

gogoyqj:
@passy,I'm soryy, we don't have an English version till now, but we are working on it.

samccone:
Hi @gogoyqj for now we are going to hold on the merge. When you have docs that are ready in english please reopen and we can evaluate. Thanks for much for the work :clap:

Issue #1084, opened 2014-12-06
Proposal: LiquidLava framework


kogarashisan:
Hello!
May I propose my own framework - LiquidLava:
http://www.lava-framework.com/

Here is the ToDo app:
http://www.lava-framework.com/todo_app/index.html#/all
Source:
https://github.com/kogarashisan/LavaSite/tree/gh-pages/todo_app

Demo passes JSHint, but it's not compatible with your unit tests and CSS,
as it uses auto-generated ID attributes on elements.
Also I prefer "is_completed" property name over "completed", and if You don't mind - I want to leave it as is.
Also, there is an issue with focus in Safari, will fix soon.

About:
I guarantee, that this framework differs a lot from anything you have seen or used before.
It was designed to cut-off development costs of large and complex web-applications,
but it's architecture easily allows it to be used for most tiny tasks and widgets.
Feature-rich and completely transparent, debuggable, easy to learn.
I offer you to have a look at tutorials on project's site.

kogarashisan:
Hello everyone!

Since this issue was opened - I continued development of my framework
- Documentation is now 95% complete (it lacks some articles about specific widgets, but core is documented well).
- Fixed some bugs in core (hope it were final bugs)
- Added some new [examples](http://www.lava-framework.com/www/examples.html#FilteredTree) and functionality

Site still lacks distinctive design (I know that default bootstrap theme looks like crap),
and there are no unit tests in the framework yet (which is compensated by numerous debug checks).

But... Lava is already powerful enough to be comparable with top of the market frameworks.
So, my question is: may I count on you and expect to be included in next update of TodoMVC?

When you are not famous - it's very hard to spread the word about your product,
no matter how good it is. Inclusion in your list would save me alot of time and efforts.

P.S.
If something depends on me - then just notify me and I will do the job.

arthurvr:
Closing for the same reason the PR was closed: https://github.com/tastejs/todomvc/pull/1133.

PR #1133, opened 2015-02-07
Added LiquidLava with NPM and new CSS heme

kogarashisan:
Added LiquidLava. with NPM and new CSS theme


examples/lava/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "private": true,
+  "dependencies": {
+    "lava": "^0.13.2",
+    "director": "~1.2.8",
+    "todomvc-app-css": "=1.0.0",
```
arthurvr: `=` -> `^`

kogarashisan: It's not a bug, it's intended.
This example does not use todomvc-app-css at all, so maybe it's better to remove it completely?

My framework is unable to work with IDs in CSS files, so I have rewritten your index.css into app.css: all ID selectors were renamed into class selectors.

For example: #todoapp -> .todoapp

And one exception:
label[for='toggle-all'] -> .toggle-all-label

P.S. Topic about the framework
https://github.com/tastejs/todomvc/issues/1084

sindresorhus:
@tastejs/todomvc

kogarashisan:
If anyone has quiestions - feel free to ask.

samccone:
Hi @kogarashisan 

Thanks for doing this great work! 
At this time we would like to see your framework mature and grow through use by more users. 
We are open to reevaluating when you feel like you have reached this point, thanks again for the great stuff.

PR #1142, opened 2015-02-09
Backbone: Clean up .gitignore

kmargaritis:


kmargaritis:
After @arthurvr commented about removing the asterisks and @passy cleaning up the .gitignore from my Flight PR i went ahead and made a PR for cleaning up Backbone .gitignore. I have searched a bit and i saw that having the asterisk actually changes the behaviour. If i add the *, git is able to see the files otherwise it does not. Could somebody clarify this. cc @sindresorhus 

P.S. I found [this](http://stackoverflow.com/questions/5533050/gitignore-exclude-folder-but-include-specific-subfolder) which points that we should put the asterisk.

sindresorhus:
Yes, the stars are necessary because of git's stupid glob handling. Can someone please add stars to where it wasn't included in the recent UI upgrades?

PR #1159, opened 2015-02-11
Knockback: upgrade to node_modules and latest library

kmalakoff:
I've upgraded the Knockback example to use node_modules as after the latest upgrade, the example was referring to an old version of Knockback.js (not from node_modules, but the lib folder).


examples/knockback/.gitignore
```diff
@@ -1,9 +1,25 @@
-node_modules/backbone/*
-!node_modules/backbone/backbone.js
+node_modules/backbone/
```
sindresorhus: ?

PR #1164, opened 2015-02-14
Update exoskeleton to 0.7.0

yavorsky:
Like my friend-bartender said: need to replace old-fashioned with something fresh. 
Actually, just updated exoskeleton to latest.

### DIFF
--- a/examples/exoskeleton/index.html
+++ b/examples/exoskeleton/index.html
@@ -53,6 +53,8 @@
 		<script src="node_modules/exoskeleton/exoskeleton.js"></script>
 		<script src="node_modules/microtemplates/index.js"></script>
 		<script src="node_modules/backbone.localstorage/backbone.localStorage.js"></script>
+		<script src="node_modules/backbone.nativeview/backbone.nativeview.js"></script>
+		<script>Backbone.View = Backbone.NativeView;</script>
 		<script src="js/models/todo.js"></script>
 		<script src="js/collections/todos.js"></script>
 		<script src="js/views/todo-view.js"></script>
--- a/examples/exoskeleton/js/views/app-view.js
+++ b/examples/exoskeleton/js/views/app-view.js
@@ -8,6 +8,12 @@ var app = app || {};
 		el.style.display = toggle ? '' : 'none';
 	};
 
+	var matchesSelector = function (node, selector) {
+		return [].some.call(document.querySelectorAll(selector), function (el) {
+			return el === node;
+		});
+	};
+
 	// The Application
 	// ---------------
 
@@ -32,10 +38,10 @@ var app = app || {};
 		// collection, when items are added or changed. Kick things off by
 		// loading any preexisting todos that might be saved in *localStorage*.
 		initialize: function () {
-			this.allCheckbox = this.find('#toggle-all');
-			this.input = this.find('#new-todo');
-			this.footer = this.find('#footer');
-			this.main = this.find('#main');
+			this.allCheckbox = this.$('#toggle-all').item(0);
+			this.input = this.$('#new-todo').item(0);
+			this.footer = this.$('#footer').item(0);
+			this.main = this.$('#main').item(0);
 
 			this.listenTo(app.todos, 'add', this.addOne);
 			this.listenTo(app.todos, 'reset', this.addAll);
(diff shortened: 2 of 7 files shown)


passy:
> Like my friend-bartender said: need to replace old-fashioned with something fresh. 



Thanks! I'll review it in a bit. :)

examples/exoskeleton/index.html
```diff
@@ -51,6 +51,8 @@
 		</script>
 		<script src="node_modules/todomvc-common/base.js"></script>
 		<script src="node_modules/exoskeleton/exoskeleton.js"></script>
+		<script src="node_modules/backbone.nativeview/backbone.nativeview.js"></script>
+		<script type="text/javascript">Backbone.View = Backbone.NativeView;</script>
```
passy: Can this be done in a script instead or is this a requirement for the dependencies that follow? In that case, could you remove the `type` attribute?

examples/exoskeleton/js/views/app-view.js
```diff
@@ -8,6 +8,15 @@ var app = app || {};
 		el.style.display = toggle ? '' : 'none';
 	};
 
+	var matchesSelector = function (el, selector) {
+		var node = el;
+		var nodes = (node.parentNode || node.document).querySelectorAll(selector);
+		var i = -1;
+
+		while (nodes[++i] && nodes[i] !== node);
+		return !!nodes[i];
```
passy: It's a neat hack, but we're trying to be as approachable as possible for newcomers. The functional alternatives are usually a bit more readable. What about this?

```js
[].slice.call(document.querySelectorAll("span")).some(function (el) { return el === node; })
```

examples/exoskeleton/js/views/app-view.js
```diff
@@ -32,10 +41,10 @@ var app = app || {};
 		// collection, when items are added or changed. Kick things off by
 		// loading any preexisting todos that might be saved in *localStorage*.
 		initialize: function () {
-			this.allCheckbox = this.find('#toggle-all');
-			this.input = this.find('#new-todo');
-			this.footer = this.find('#footer');
-			this.main = this.find('#main');
+			this.allCheckbox = this.$('#toggle-all')[0];
```
passy: That looks like a regression, API-wise? Is there no longer a way to target a single node?

yavorsky: @passy it returns NodeList, so we can use `this.$('#toggle-all').item(0)` or `this.el.querySelector('#toggle-all')`. According to exoskeleton [changelog](https://github.com/paulmillr/exoskeleton/blob/master/CHANGELOG.md), View#find was removed since 0.7.0.

paulmillr: yeah, meh. we should talk about this in exos issue tracker

examples/exoskeleton/js/views/app-view.js
```diff
@@ -8,6 +8,12 @@ var app = app || {};
 		el.style.display = toggle ? '' : 'none';
 	};
 
+	var matchesSelector = function (node, selector) {
+		return [].some.call(document.querySelectorAll(selector), function (el) {
```
samccone: `Array.prototype.some` then no need to allocate a new array instance just to use the method eh :)

examples/exoskeleton/js/views/app-view.js
```diff
@@ -66,9 +72,9 @@ var app = app || {};
 					remaining: remaining
 				});
 
-				this.findAll('#filters li a').forEach(function (el) {
+				[].forEach.call(this.$('#filters li a'), function (el) {
```
samccone: ditto ^

samccone:
some tiny nits, and we should run the test suite over this, but I am :+1:

examples/exoskeleton/js/views/app-view.js
sindresorhus: @samccone JS engines optimizes this away, so doesn't matter.

samccone:
I am getting three CI failures on this branch 

```
     1) should display the number of completed items
      2) should remove completed items when clicked
      3) should be hidden when there are no items that are completed
```

Can someone else verify?

samccone:
ok looks like those three tests are all broken with the new UI :crying_cat_face: working on a fix upstream
thanks for the work this is going in :rugby_football:

PR #1166, opened 2015-02-16
Exoskeleton: Switch to the original backbone.localstorage

yavorsky:


sindresorhus:
Can you update the actual file too? Or it identical?

yavorsky:
@sindresorhus Identical. Difference only in `package.json`.

Issue #789, opened 2014-01-23
Editing is not cancelled on ESCAPE keypress


ColinEberhardt:
The specification states that edits should be cancelled on escape:

> If escape is pressed during the edit, the edit state should be left and any changes be discarded.

The current implementations do not meet this requirement:
- [x] agilityjs
- [x] angularjs-perf
- [x] thorax_lumbar
- [x] flight
- [x] backbone_require
- [x] knockoutjs_require
- [x] closure
- [x] jquery
- [x] knockback
- [x] spine
- [x] ariatemplates
- [x] batman
- [x] ~~dermis~~
- [ ] dijon
- [ ] duel
- [ ] extjs_deftjs
- [x] kendo
- [x] olives
- [x] plastronjs
- [x] rappidjs
- [ ] serenadejs
- [x] somajs
- [x] thorax
- [x] typescript-angular
- [x] angularjs_require
- [x] backbone_marionette_require

arthurvr:
I opened #1223.

arthurvr:
pinging @podefr for the olives app. You interested in taking this one?

podefr:
@arthurvr of course, will do, just need to other code review to go in first to build on solid grounds :)

podefr:
@arthurvr I just created a PR to fix this issue #1287

dprotti:
Closure impl seems to be working fine http://todomvc.com/examples/closure/#/

arthurvr:
@dprotti Yep. Tests pass as well. Thanks for letting us know. I guess it was fixed but we just never ticked it on the list. Done now.

arthurvr:
Pinging in the app authors and frequent contributors to see if we can push this issue forward. @eastridge @mkuklis @margaritis @loicknuchel @creynders @mckamey @brian428 @jnicklas Can y'all take a look?

mckamey:
Sending a PR shortly.

PR #1168, opened 2015-02-17
Fix spec compliance for Marionette.

samccone:
#### What:

In issue #789 @ColinEberhardt pointed out that many of the examples do
not respect the rule that when a user hits the esc key on an input that
the input should be cleared and the input state should be left.

Upon investigation I found the Marionette example to have never followed
this directive.

#### How:

Due the how the escape key works and the keypress event binding from
jQuery one must use the keyup event to register for the escape key
keycode.

Once I intercept the event I instruct the input area to simply rerender.
This has the desired effect in that it clears the transient input state
and also due to it rendering new childnodes clears the focus from the
user.

------

cc @ColinEberhardt

### DIFF
--- a/examples/backbone_marionette/js/TodoMVC.Layout.js
+++ b/examples/backbone_marionette/js/TodoMVC.Layout.js
@@ -14,7 +14,18 @@ TodoMVC.module('Layout', function (Layout, App, Backbone) {
 		},
 
 		events: {
-			'keypress @ui.input': 'onInputKeypress'
+			'keypress @ui.input': 'onInputKeypress',
+			'keyup @ui.input': 'onInputKeyup'
+		},
+
+		// According to the spec
+		// If escape is pressed during the edit, the edit state should be left and any changes be discarded.
+		onInputKeyup: function (e) {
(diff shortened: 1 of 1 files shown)


examples/backbone_marionette/js/TodoMVC.Layout.js
```diff
@@ -14,7 +14,18 @@ TodoMVC.module('Layout', function (Layout, App, Backbone) {
...
+			'keypress @ui.input': 'onInputKeypress',
+			'keyup @ui.input': 'onInputKeyup'
+		},
+
+		// According to the spec
+		// If escape is pressed during the edit, the edit state should be left and any changes be discarded.
+		onInputKeyup: function (e) {
+			var ESC_KEY = 27;
+
+			if (e.keyCode == ESC_KEY) {
```
sindresorhus: Since jQuery is used `e.which` is probably better.

Also, use `===`.

samccone:
@sindresorhus thank you for the review, and the spot. Addressed and resolved.

Issue #1110, opened 2015-01-30
New app UI & switch to npm for dependencies


sindresorhus:
I finally managed to finish the [new app UI](https://github.com/tastejs/todomvc-app-css).

Now we just need to switch the apps over. No structural changes should be needed.

I took the liberty to change from bower to npm as that's the future and it makes it easier for us to deal with dependencies. We should also .gitignore any unwanted file so to make it easier to upgrade dependencies in the future.

`todomvc-common` no longer contains the app CSS, only the learning-bar. The app CSS is in the `todomvc-app-css` module.

I've converted the [backbone app as an example](https://github.com/tastejs/todomvc/commit/d674391ea8334ee663b6b050b6b88863a3cba90f). ([live demo](http://todomvc.com/examples/backbone/)).
## 

Help from anyone more than welcome.
## Apps needing updating
- [x] FOAM
- [x] agilityjs (dropped, https://github.com/tastejs/todomvc/issues/1173)
- [x] ampersand
- [x] angular-dart
- [x] angularjs
- [x] angularjs_require
- [x] angularjs-perf
- [x] ariatemplates
- [x] atmajs
- [x] backbone
- [x] backbone_marionette
- [x] backbone_require
- [x] batman
- [x] canjs
- [x] canjs_require
- [x] chaplin-brunch
- [x] closure
- [x] componentjs
- [ ] cujo
- [x] derby
- [x] dijon
- [x] dojo
- [ ] duel
- [ ] durandal
- [x] emberjs
- [x] enyo_backbone
- [x] epitome
- [x] exoskeleton
- [x] extjs_deftjs
- [x] firebase-angular
- [x] flight
- [ ] gwt
- [x] jquery
- [ ] kendo
- [x] knockback
- [x] knockoutjs
- [x] knockoutjs_require
- [x] lavaca_require
- [x] maria (dropped, https://github.com/tastejs/todomvc/issues/1173)
- [ ] meteor
- [x] mithril
- [x] montage
- [x] olives
- [x] plastronjs
- [x] polymer
- [x] puremvc
- [x] ractive
- [x] rappidjs
- [x] react
- [x] react-backbone
- [ ] sammyjs
- [ ] sapui5
- [x] serenadejs
- [ ] socketstream
- [x] somajs
- [x] somajs_require
- [x] spine
- [x] stapes
- [x] stapes_require
- [ ] troopjs_require
- [x] typescript-angular
- [x] typescript-backbone
- [x] vanilladart
[...]

PR #1169, opened 2015-02-17
Sjs/update agility app ui

samccone:
This project does not have a package.json and has not been active since
June 1 2013, so instead of waiting for a package.json to be merged I
choose to use a project called napa to allow for installs of packages
without a package.json

-------

Via #1110


sindresorhus:
> This project does not have a package.json and has not been active since June 1 2013

Oh wow. That's a pretty good sign we should remove it from TodoMVC altogether. We can put a link to your version in the wiki. @tastejs/todomvc thoughts?

samccone:
After all my work :( bahahahah

samccone:
but yes :+1: to remove

sindresorhus:
@samccone Yeah, sorry about that. I had no idea idea it was that dead... But we're not doing our users any favor of having it there when the project is dead.

addyosmani:
:+1: to removing but thanks for working on this @samccone. My vote is to link to this from our wiki.

sindresorhus:
@samccone Can you update the wiki entry to point to your fork?

https://github.com/tastejs/todomvc/wiki/Other-implementations#agility-unmaintained

PR #1183, opened 2015-03-02
Streamline installing/running the browser tests

chrisprice:
Whilst updating the closure app, I found running the browser tests a bit cumbersome. I've updated them to use npm scripts rather than global installations or grunt.

I kept the ```serve``` script separate from the ```test``` to keep the commands cross-platform.


passy:
That is awesome, thank you! Just tried and it works a treat. It's great no longer having to have a global grunt-cli installation for this.

PR #1184, opened 2015-03-06
Remove deprecated global addRegion in Marionette App

### LINKED ISSUE #1171: Marionette Application Region to be Deprecated
````
This feature is deprecated. Instead of using the Application as the root of your view tree, you should use a Layout View. To scope your Layout View to the entire document, you could set its el to 'body'. 
````

I think the use of `addRegion` should be replace as per the docs says.

HenriqueSilverio:
Hi guys, I'm looking for update the `backbone_marionette` example.

This pull request attempts to resolve #1171, updating Marionette version, and using the concept of a [App.rootEl](https://github.com/marionettejs/backbone.marionette/issues/2279) instead "global" regions direct on Application instance. Any suggestions? =]

Thanks!

### DIFF
--- a/examples/backbone_marionette/js/TodoMVC.Layout.js
+++ b/examples/backbone_marionette/js/TodoMVC.Layout.js
@@ -2,6 +2,18 @@
 'use strict';
 
 TodoMVC.module('Layout', function (Layout, App, Backbone) {
+	// Layout Root View
+	// ------------------
+	Layout.Root = Backbone.Marionette.LayoutView.extend({
+		el: 'body',
+
+		regions: {
+			header: '#header',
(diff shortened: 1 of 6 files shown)


sindresorhus:
// @samccone

samccone:
Oh yes we are on it 😊 I am but one man in a sea of many 😉

1vsM:
One vs Many! salute



*One vs Many [**Fashion | Events | Music | Community]*
*www.ONEVSMANY.com <http://onevsmany.com/> | *77-HAPPY-202 |
774-277-9202 | @ONEVERSUSMANY
#1vsM

PR #1185, opened 2015-03-08
Update Marionette & Remove Deprecated Feature use

### LINKED ISSUE #1171: Marionette Application Region to be Deprecated

samccone:
Via #1184 -- Thanks @HenriqueSilverio 

Marionette has marked `appRegions` as a deprecated feature because it is a
leaky abstraction for the Application to know so much about the DOM and
the structure of your actual page.
Instead of using `appRegions`, Marionette encourages you to setup a
RootLayout whose concern is setting up render targets.

In the near future we are introducing the concept of a "root" to the
application object

marionettejs/backbone.marionette#2279

So that uses will have an idiomatic way of using this pattern. The goal
here is not to conflate the concerns of splitting up render targets
with your application, rather the goal is to give users a pattern to follow when
setting up a single or multiple application objects.

Fixes #1171
Fixes #1184

### DIFF
--- a/examples/backbone_marionette/js/TodoMVC.Layout.js
+++ b/examples/backbone_marionette/js/TodoMVC.Layout.js
@@ -2,6 +2,16 @@
 'use strict';
 
 TodoMVC.module('Layout', function (Layout, App, Backbone) {
+
+	Layout.Root = Backbone.Marionette.LayoutView.extend({
+		el: '#todoapp',
+		regions: {
+			header: '#header',
+			main: '#main',
+			footer: '#footer'
+		}
+	});
+
 	// Layout Header View
 	// ------------------
 	Layout.Header = Backbone.Marionette.ItemView.extend({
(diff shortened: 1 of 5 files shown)


addyosmani:
Visually verifying the app looks as expected:

screen shot 2015-03-08 at 19 08 00

addyosmani:
We appreciate you updating this implementation @samccone. Code reviewed, but didn't find anything out of place that caught my eye. Merging away! :+1:

HenriqueSilverio:
Nice implementation, thanks @samccone. =]

samccone:
@HenriqueSilverio Very close to yours even when I did it in isolation 👏 thanks so much for the great work

PR #1199, opened 2015-03-13
Center footer credits

samccone:
I think this looks much nicer given the new constraints.

##### Before
screen shot 2015-03-12 at 11 40 01 pm

##### After
screen shot 2015-03-12 at 11 40 08 pm


sindresorhus:
Yeah, I also think we should put the `Brought to you by` on a separate line above our names.

PR #1206, opened 2015-03-15
Dart update

MathieuLorber:
A vanilladart update.

I removed the uuid library which actually isn't intended to be used in the browser - the JS compilation result is now 137ko (instead of 529). I'm looking into the possibility of a uuid lib dedicated to the browser, will make another update if so.

BTW, a new Dart formatter is available, and the global formatting of the example isn't consistent. Should I make a formatting pass on all the files ?


sindresorhus:
> Should I make a formatting pass on all the files ?

Yes, but in a separate PR.

MathieuLorber:
>> Should I make a formatting pass on all the files ?

> Yes, but in a separate PR.

Sure

sindresorhus:
Thanks @MathieuLorber :)

MathieuLorber:
Again, I wish I could be more reactive...

cc @passy , need help for an update of the angulardart sample ?

MathieuLorber:
For the record, I just realized the uuid lib in current version + the current Dart compiler actually produce a 149ko JS file. But let's keep the "inline implementation" for the moment...

PR #1210, opened 2015-03-16
Run a subset of the tests on every PR!!!

### LINKED ISSUE #946: Run automated tests with Travis
Useful to see whether new PRs passes the tests or not.

@colineberhardt would you be able to do this?

samccone:
Master still runs deploy but all the other than that the following happens

* First we add the real remote and fetch the contents
* then we diff the current branch from masters contents
* then we get the examples folders.. for instance

examples/backbone/....
examples/angular/.....

then I awk out the middle name, which happens to be used with our test
suite to run each test. From there I use xargs to invoke the tests
running over a subset of all of the tests.

```shell
npm run test -- --framework=backbone --framework=angular
--framework=dogescript
```

---

commit 1 allows us to do this

`npm run test -- --framework=jquery --framework=backbone --grep 'should trim entered text'`

------

Fixes #946

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,6 @@ node_js:
     - '0.12'
 notifications:
     irc: 'irc.freenode.org#tastejs'
-branches:
-    only:
-        - master
 env:
     global:
         # GH_OAUTH_TOKEN is the oauth token generated as described at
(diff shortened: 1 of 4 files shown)


samccone:
~~~ tweaking travis atm ~~~

Issue #1216, opened 2015-03-17
Policy about separated `require.js` apps?


arthurvr:
Discussion started here: https://github.com/tastejs/todomvc/commit/79c34b30006f91f35b14295087bcc477ba8ea990#commitcomment-10202521. We already decided to remove `backbone-marionette-requirejs` (https://github.com/tastejs/todomvc/commit/79c34b30006f91f35b14295087bcc477ba8ea990).

> the backbone marionette require version can be removed, i don't think module systems are new enough anymore to warrant a separate version.

-

We have quite a bunch of these `require.js` apps, and are gonna stay getting more and more. It's a PITA to maintain two different versions of those apps. Personally I also feel like module systems aren't new enough anymore to have separated apps for them.

Apps that fall under this:

* angularjs_require
* backbone_require
* canjs_require
* emberjs_require
* knockoutjs_require
* somajs_require

samccone:
:+1: to dropping

arthurvr:
I'm also a huge +1 to dropping.

sindresorhus:
Cyclopean :+1:

I was always against this.

mikaelkaron:
Not that it matters but I'm +1 on dropping to (the only reason I care is that that would hopefully rename `troopjs_require` to simply `troopjs`)

arthurvr:
The troopjs example isn't gonna be effected by this, as there is no duplicating, there is only one app.

> that would hopefully rename troopjs_require to simply troopjs

Maybe we should indeed just rename it. @tastejs/todomvc Thoughts?

addyosmani:
I'm :+1: on renaming the Troop example to just TroopJS. Definitely avoids confusion - good idea @arthurvr! :star:

PR #1220, opened 2015-03-19
Drop `require.js` apps

arthurvr:
Looks like we're all alright with this: https://github.com/tastejs/todomvc/issues/1216.

Fixes https://github.com/tastejs/todomvc/issues/1216


addyosmani:
I'm -1 on dropping these implementations, but would be open to considering given more data. I do appreciate the conversation around this being started though. Thanks @arthurvr :)

Some questions:

* When was the last time the above implementations were updated?
* How difficult have we found it to get these implementations updated after having the main, non RJS version updated?
* How do the analytics usage stats for these apps compare to the others in the collection?

arthurvr:
> I'm -1 on dropping these implementations given AMD is still heavily used by those who haven't switched to CJS or ES6.

Still, I personally don't think the specific AMD examples add many value. People that are used to using AMD don't need examples on how to use it anymore, IMHO. If they're wanting to explore a new framework, I don't think it adds much value for them to see it written using AMD.

> When was the last time the above implementations were updated?

For all of them I think we can state the activity is fairly (/really) low. These get outdated with the 'main' example way to quick, which is a result of the duplicating we're doing.

* `canjs_require`: https://github.com/tastejs/todomvc/commits/master/examples/canjs_require
* `backbone_require`: https://github.com/tastejs/todomvc/commits/master/examples/backbone_require
* `angularjs_require`: https://github.com/tastejs/todomvc/commits/master/examples/angularjs_require
* `emberjs_require`: https://github.com/tastejs/todomvc/commits/master/examples/emberjs_require
* `knockoutjs_require`: https://github.com/tastejs/todomvc/commits/master/examples/knockoutjs_require
* `somajs_require`: https://github.com/tastejs/todomvc/commits/master/examples/somajs_require

> How difficult have we found it to get these implementations updated after having the main, non RJS version updated?

Hard, I guess. The amount [of known issues with those](https://github.com/tastejs/todomvc/pull/1220/files#diff-8b25e71f0d3d34cd7c1fe864bf1f6ded) is quite big.

passy:
We've had a pretty long discussion about this for the AngularJS example and decided to provide one because there was a pretty substantial need from the community for a reference implementation.

That example is in fact still our 8th most popular example. So I'm definitely :-1: for on this, for now. I think we all agree that it would be nice in the long run to remove these.

screenshot from 2015-03-19 20 59 52

addyosmani:
I agree that in the long term we should definitely seek to drop. I'd actually far prefer to see Browserify or ES6 examples at some point, but for now, I think it makes sense to keep them around.

arthurvr:
The analytics convinced me, though I guess for the other examples it does make less sense to have this duplicative implementations. Closing for now.

PR #1221, opened 2015-03-20
Delegate events to #todo-list instead of document

evverx:


evverx:
This change inspired from [.live() vs .delegate()](http://www.elijahmanor.com/differences-between-jquery-bind-vs-live-vs-delegate-vs-on/) battle.
`live` cons:
* All your events always delegate all the way up to the document.

@cburgmer , what do you think?

passy:
That looks reasonable to me. Also, shorter!

cburgmer:
Like it!

samccone:
heh :+1:

arthurvr:
I guess this one is good to go? :+1: of mine

PR #1223, opened 2015-03-21
Revert on escape for `angularjs_require` app

arthurvr:
Ref #789. Tests passing.

### DIFF
--- a/examples/angularjs_require/js/controllers/todo.js
+++ b/examples/angularjs_require/js/controllers/todo.js
@@ -52,6 +52,8 @@ define(['app', 'services/todoStorage'], function (app) {
 
 			$scope.editTodo = function (todo) {
 				$scope.editedTodo = todo;
+				// Clone the original todo to restore it on demand.
+				$scope.originalTodo = angular.copy(todo);
 			};
 
 
@@ -64,6 +66,10 @@ define(['app', 'services/todoStorage'], function (app) {
 				}
 			};
 
+			$scope.revertEditing = function (todo) {
+				todos[todos.indexOf(todo)] = $scope.originalTodo;
+				$scope.doneEditing($scope.originalTodo);
+			};
 
 			$scope.removeTodo = function (todo) {
 				todos.splice(todos.indexOf(todo), 1);
--- a/examples/angularjs_require/js/directives/todoEscape.js
+++ b/examples/angularjs_require/js/directives/todoEscape.js
@@ -1,0 +1,16 @@
+/*global define*/
+define(['app'], function (app) {
+	'use strict';
+
+	app.directive('todoEscape', function () {
+		var ESCAPE_KEY = 27;
+
+		return function (scope, elem, attrs) {
+			elem.bind('keydown', function (event) {
+				if (event.keyCode === ESCAPE_KEY) {
+					scope.$apply(attrs.todoEscape);
+				}
+			});
+		};
+	});
+});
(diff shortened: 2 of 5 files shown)


arthurvr:
@tastejs/todomvc One wanting to give this a review or shall I go ahead?

examples/angularjs_require/js/directives/todoEscape.js
```diff
@@ -0,0 +1,16 @@
+/*global define*/
+define(['app'], function (app) {
+	'use strict';
+
+	app.directive('todoEscape', function () {
+		var ESCAPE_KEY = 27;
+
+		return function (scope, elem, attrs) {
+			elem.bind('keydown', function (event) {
+				if (event.keyCode === ESCAPE_KEY) {
```
samccone: since we are setting this binding outside of angular's knowledge we are going to want to clean it up via 

`scope.$on('$destroy', function() {`

otherwise we are going to get into leak hell

unlessssss
https://docs.angularjs.org/api/ng/function/angular.element#events

is magically cleaning this up for us, thoughts @arthurvr ?

arthurvr: I honestly have no idea if it is taking care of that. cc'ing @passy as this was mostly copy-pasta from https://github.com/tastejs/todomvc/commit/2f6242683818611a585ff502671db08c5af06d6f.

examples/angularjs_require/js/controllers/todo.js
```diff
@@ -52,6 +52,8 @@ define(['app', 'services/todoStorage'], function (app) {
 
 			$scope.editTodo = function (todo) {
 				$scope.editedTodo = todo;
+				// Clone the original todo to restore it on demand.
+				$scope.originalTodo = angular.extend({}, todo);
```
passy: You can use `angular.copy` for this

examples/angularjs_require/js/controllers/todo.js
```diff
@@ -64,6 +66,11 @@ define(['app', 'services/todoStorage'], function (app) {
 				}
 			};
 
+			$scope.revertEditing = function (todo) {
+				todos[todos.indexOf(todo)] = $scope.originalTodo;
+				$scope.doneEditing($scope.originalTodo);
+			};
+
```
passy: Minor: extra line

examples/angularjs_require/js/directives/todoEscape.js
passy: Good catch, Sam! Can we somehow verify if the bindings leak?

samccone: I think a  snapshot would be the easiest way...

examples/angularjs_require/js/controllers/todo.js
arthurvr: It's like this in other apps as well, but changing. https://github.com/tastejs/todomvc/blob/master/examples/angularjs/js/controllers/todoCtrl.js#L55

examples/angularjs_require/js/controllers/todo.js
arthurvr: kk ;)

examples/angularjs_require/js/controllers/todo.js
arthurvr: @passy All other angular examples do it this way so shall I fix it up in all of them?

examples/angularjs_require/js/directives/todoEscape.js
arthurvr: @samccone @passy This is the way other apps [are doing it as well](https://github.com/tastejs/todomvc/blob/master/examples/angularjs/js/directives/todoEscape.js#L18) so maybe we should separate this out in another issue and go forward on this PR?

samccone: obviously there is a bunch of copypasta going on here :) but I would rather not introduce something broken right?

samccone: (just because it is broken in other places does not make it ok to merge broken things)

samccone:
:rocket:

Issue #1215, opened 2015-03-17
Clear completed integration tests are failing as a result of the new UI


samccone:
since we now set `clear completed` in CSS https://github.com/tastejs/todomvc-app-css/blob/master/index.css#L325 the integration tests that we have no longer pass. 

This is unfortunate because it impacts any app that has upgraded to the new UI. We should investigate how we would like to resolve it, since it would be great to get a passing suite once more.

:+1: :sparkles: :spaghetti:

evverx:
`clear completed` in CSS breaks Firefox: [Firefox: Click on CSS pseudo element doesn't trigger click event](http://stackoverflow.com/questions/21858942/firefox-click-on-css-pseudo-element-doesnt-trigger-click-event).
IE11 doesn't show button at all
![](https://leto13d.storage.yandex.net/rdisk/4203feb601316ef790830a5d67db152faf6281d83a1f3078dd6f52d10046259d/inf/-DDIprZESaDwZiD2vmMTyP9jZsi2c0xSPU7dGe4iReJ95UHO5BslnUR4SLTm_7jvWkbpNvr8MYSHkj3mtQ9_iQ==?uid=0&filename=2015-03-20%2007-54-31%20IE11%20-%20Win7%20%5BRunning%5D.png&disposition=inline&hash=&limit=0&content_type=image%2Fpng&tknv=v2&rtoken=f263067d0f905c659bb803d917f2a372&force_default=no)

arthurvr:
I'll hopefully find some time to look into this.

samccone:
Fixed via 
https://github.com/tastejs/todomvc-app-css/pull/7

PR #1228, opened 2015-03-23
Update index.html

elie29:
We should display the completed todos after the clear link

### DIFF
--- a/examples/backbone/index.html
+++ b/examples/backbone/index.html
@@ -46,7 +46,7 @@
 				</li>
 			</ul>
 			<% if (completed) { %>
-			<button id="clear-completed">Clear completed (<%= completed %>)</button>
+			<button id="clear-completed">Clear completed (<%- completed %>)</button>
 			<% } %>
 		</script>
 		<script src="node_modules/todomvc-common/base.js"></script>


samccone:
:+1: to escaping, not sure how this lines up with your PR message though?

elie29:
What do you mean?

samccone:
> We should display the completed todos after the clear link

not sure how this lines up the the change to interpolating

elie29:
http://todomvc.com/examples/backbone/
When you check a todo, "clear completed" is displayed without the number of completed todos. But when you test with my commit, you will get the correct result : 
image

elie29:
I should put this message : displaying the number of completed to using minus sign for HTML-escaped (cf. underscore)

sindresorhus:
:+1: but please include a proper commit and PR title.

arthurvr:
> I should put this message : displaying the number of completed to using minus sign for HTML-escaped (cf. underscore)

I don't know if you actually tested this, but it doesn't work. Currently this worked fine, but we were just running into https://github.com/tastejs/todomvc/issues/1215. The actual problem will get fixed when we fix todomvc-app-css, this does not fix anything.

Do we still need these PRs, then?

elie29:
No it is not necessary :+1:

PR #1233, opened 2015-03-25
Fix minor style nits

samccone:
There were some style errors in the test file, this fixes them

### DIFF
--- a/browser-tests/page.js
+++ b/browser-tests/page.js
@@ -25,7 +25,7 @@ module.exports = function Page(browser) {
 
 	// ----------------- navigation methods
 
-	this.back = function() {
+	this.back = function () {
 		return browser.navigate().back();
 	};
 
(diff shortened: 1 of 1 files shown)


browser-tests/page.js
```diff
@@ -152,7 +152,9 @@ module.exports = function Page(browser) {
 		return this.getItemLabelAtIndex(index).then(function (itemLabel) {
 			// double click is not 'natively' supported, so we need to send the
 			// event direct to the element see:
+			// jscs:disable
```
arthurvr: why? If jscs is complaining about this that's a jscs bug.

samccone: this is too long of a line so it complains this disables it for this line

arthurvr: Oh that way. Yup.

PR #1235, opened 2015-03-26
Fix getting started with Dojo link

peterkokot:
Current link points to missing page. If I may suggest to include the getting started tutorial from Dojo. Thank you.


arthurvr:
Looks good. Thanks.

PR #1238, opened 2015-03-28
Remove React onEdit() workaround

leftdevel:
- New Feature: No.
- BugFix: No.
- Refactor: Yes.

Hi there.
Before anything I'd like to say thanks for this amazing project, it helped a lot to decide on a JS library.

I revisited the React code today and I noticed a workaround in TodoApp::onEdit() and in TodoItem::handleEdit() functions.

With this pull request I'm aiming to simplify the code  by moving the DOM manipulation to a more appropriated place where it feels more natural  to the React lifecycles.

We will also make the TodoApp crud methods more consistent. I noticed that only onEdit() function expects a callback to be passed, and it feels harder to reason about.

If you noticed I moved the DOM manipulation to `componentDidUpdate` method. While this make things feel "more natural", I think it also promotes a switch-case design which is not good either.

However I find more advantages by doing it this way, like, for example, that now the node.focus() logic stays encapsulated within the TodoItem and we avoid making the TodoApp aware that it needs to execute a callback whenever onEdit is invoked.

What do you guys think? Regards.

### DIFF
--- a/examples/react/js/app.jsx
+++ b/examples/react/js/app.jsx
@@ -62,12 +62,8 @@ var app = app || {};
 			this.props.model.destroy(todo);
 		},
 
-		edit: function (todo, callback) {
-			// refer to todoItem.js `handleEdit` for the reasoning behind the
-			// callback
-			this.setState({editing: todo.id}, function () {
-				callback();
-			});
+		edit: function (todo) {
+			this.setState({editing: todo.id});
 		},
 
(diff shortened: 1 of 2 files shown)


passy:
That looks great. Moving the DOM manipulation to the mount handlers seems like the right thing to do. I want to do some manual testing to be safe, but thanks in advance for the contribution.

passy:
Works a treat. Thank you!

PR #1239, opened 2015-03-29
Fix id -> class in index.html

JacobDanton:
The current index.html incorrectly uses "id" in place of "class" and does not match the styling in todomvc-app-css


arthurvr:
That's the reason we still use todomvc-app-css 1.0.1. We're planning to change them over to classes but the unit tests will first need to get updated (https://github.com/tastejs/todomvc-app-css/issues/4 FYI)

arthurvr:
We're gonna do all examples in one pass so closing this for now. Thanks, though.

Issue #1248, opened 2015-04-03
Memory leak in yui example


masrud:
Hi there,
The Todo example developed using YUI library generates leaked objects. After adding multiple items to the todo list and then removing them, it is expected that the allocated memory is released. However, I observer an increasing trend in the memory use of the page as shown by Timeline memory tab.

arthurvr:
Closing as we merged the PR ^

PR #1249, opened 2015-04-03
Fixing a memory leak in yui example

masrud:
Hi,

This pull request is to fix a memory leak in yui example, issue #1248.
The leak is caused because the view objects are not completely destroyed.

Thanks,
Masrud

### DIFF
--- a/examples/yui/js/app.js
+++ b/examples/yui/js/app.js
@@ -115,6 +115,9 @@ YUI.add('todo-app', function (Y) {
 				fragment.append(view.render().get('container'));
 			});
 
+			// To delete previously attached views and fix the memory leak.
+			this.get('container').one('#todo-list').destroy({recursivePurge: true});
+
 			this.get('container').one('#todo-list').setContent(fragment);
 		},
 


examples/yui/js/app.js
```diff
@@ -115,6 +115,9 @@ YUI.add('todo-app', function (Y) {
 				fragment.append(view.render().get('container'));
 			});
 
+			// To delete previously attached views and fix the memory leak.
+			this.get('container').one('#todo-list').destroy({recursivePurge: true});
```
samccone: what a great API flag

`recursivePurge`

samccone:
hey @masrud this looks correct to me, but would you mind running a before and after snapshot of the DOM nodes?

masrud:
Hi Sam,

yes, just calling destroy does not fix the leak and the flag is required.

I attached a figure of the memory use of the page before and after fixing
the leak. Memory is measured using "capture memory" in Timeline tab of the
developer tools.

In each case, I did the following activities on a page:

for three times
   added 8 tasks, removed 8 tasks, performed a garbage collection

The first figure (yuibefore.png) is generated before fixing the leak, as
can be seen, there is an increasing trend.

The second figure (yuiAfterAddingDestroy.png) is generated after making the
changes in the code (calling destroy function). This fix removes one memory
leak.

There is still another memory leak in the page. This memory leak is caused
by lines 76-80 of the "app.js" file. To fix it, temporarily I commented the
lines.
The third figure (yuiAfterfixingAnotherLeak.png) shows the memory use after
this last fix.

Bests,
Masrud

masrud:
Hi Sam,

yes, just calling destroy does not fix the leak and the flag is required.

I attached a figure of the memory use of the page before and after fixing the leak. Memory is measured using "capture memory" in Timeline tab of the developer tools.

In each case, I did the following activities on a page:

for three times
   added 8 tasks, removed 8 tasks, performed a garbage collection

The first figure (yuibefore.png) is generated before fixing the leak, as can be seen, there is an increasing trend.

The second figure (yuiAfterAddingDestroy.png) is generated after making the changes in the code (calling destroy function). This fix removes one memory leak.

There is still another memory leak in the page. This memory leak is caused by lines 76-80 of the "app.js" file. To fix it, temporarily I commented the lines.
The third figure (yuiAfterfixingAnotherLeak.png) shows the memory use after this last fix.

Bests,
Masrud

yuibefore
yuiafteraddingdestroy
yuiafterfixinganotherleak

samccone:
:clap: lgtm anyone else want to take a peak?

addyosmani:
This change LGTM (also happy to see more use of profiling tools to nail down our memory leaks). Landing!

PR #1258, opened 2015-04-07

elie29:
If we don't debounce "this.render", the app-view will be rendered 4 times on each todo (visible, change, change:complete and sync). So when we try to check globally all todos, this.render is called a hundred times. Once this PR is accepted, I will edit Backbone-requirejs.

Another problem I have addressed concerning the global checkbox. It should be checked when we have at least one completed todo.

### DIFF
--- a/examples/backbone/js/views/app-view.js
+++ b/examples/backbone/js/views/app-view.js
@@ -38,7 +38,7 @@ var app = app || {};
 			this.listenTo(app.todos, 'reset', this.addAll);
 			this.listenTo(app.todos, 'change:completed', this.filterOne);
 			this.listenTo(app.todos, 'filter', this.filterAll);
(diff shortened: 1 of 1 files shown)


examples/backbone/js/views/app-view.js
```diff
@@ -70,7 +71,7 @@ var app = app || {};
 				this.$footer.hide();
 			}
 
-			this.allCheckbox.checked = !remaining;
+			this.allCheckbox.checked = size && !remaining;
```
elie29: The checkbox is checked only if one completed todo exists

samccone: I am not able to ever hit this new check.. can you drop this change and open a PR with only the debounce change and we can get this in. thanks again @elie29

PR #1263, opened 2015-04-12
Implements Node on Fire example

martijndeh:
This implements TodoMVC in Node on Fire. A full-stack isomorphic web framework built on top of Node.js, AngularJS, Express and PostgreSQL.

The example should abide to the app specification. I also added a small fix to hide selection when double clicking a todo item.

Please let me know if I need to make any changes and/or how to continue.


samccone:
Hi @martijndeh this looks like a neat new project! 

First off we typically want to see a community around a new "framework" before it is added to the site. IT seems like this framework is brand new and we would love to see it develop more and build a user base before putting it on the site.

When you feel like it has reached this point, please feel free to reopen and we can evaluate at that time.

Thanks
Sam

PR #1271, opened 2015-04-25
Resolves #1270 by moving backbone.history.start invocation

### LINKED ISSUE #1270: Backbone.Marionette router history placement?
Hi there,

I've just been going through the source of the example.

However, I was wondering about `Backbone.history.start()` placement.

If a user interacts with the application and then tries to copy the url into another browser tab (bookmarking), route handlers won't be fired. It isn't too important for the example but where the `TodoList.Controller` is instantiated, if you place `Backbone.history.start()` here before the controller is started, the route handling will be respected.

I don't think binding to application start respects an order so one is getting called before the other and therefore the controllers routes aren't respected.

@jh3y

jh3y:


samccone:
:+1: thanks for the fast fix, nice work

jh3y:
No problem! :+1:

arthurvr:
Nice work! One nitpick:

> When using anonymous functions, leave a space between the function name and opening parenthesis.

From the [_code style guide_](https://github.com/tastejs/todomvc/blob/master/codestyle.md#anonymous-functions). Fixed in https://github.com/tastejs/todomvc/commit/65a8ecfae02242dc864ea8c956a8ecbfc41b9c11.

jh3y:
My bad, not sure how I missed that :laughing:

Thanks for pointing that out @arthurvr !

Issue #834, opened 2014-02-05
Browser tests do not verify that todo items are persisted


ColinEberhardt:
A test for this exists, but has been commented out because it fails intermittently:

https://github.com/tastejs/todomvc/blob/gh-pages/browser-tests/test.js#L286

PR #1212, opened 2015-03-17
CI Test work

samccone:
Master still runs deploy but all other branches and PR's will run the tests on the changes.

:sparkles: :dancer: :sparkles: 

I anticipate a bit of friction when we flip this on.. just with the sauce keys and such but, everything works on my fork :+1: 

* First we add the real remote and fetch the contents
* then we diff the current branch from masters contents
* then we get the examples folders.. for instance

examples/backbone/....
examples/angular/.....

then I awk out the middle name, which happens to be used with our test
suite to run each test. From there I use xargs to invoke the tests
running over a subset of all of the tests.

```shell
npm run test -- --framework=backbone --framework=angular
--framework=dogescript
```

---

commit 1 allows us to do this

`npm run test -- --framework=jquery --framework=backbone --grep 'should trim entered text'`

------

Fixes #946

PR #1272, opened 2015-04-26
Tests: add back `persistence` test

arthurvr:
In the past we commented out this test because it failed intermittently. This was simply because the `otherUrl` variable didn't exist and an error was thrown when trying to access it.

Changing `otherUrl` with `about:blank` seemed to fix the issue. [This should, normally, work cross-browser.](http://en.wikipedia.org/wiki/About_URI_scheme)

@ColinEberhardt Would love to hear your thoughts on this.

Fixes https://github.com/tastejs/todomvc/issues/834

### DIFF
--- a/browser-tests/test.js
+++ b/browser-tests/test.js
@@ -285,7 +285,7 @@ module.exports.todoMVCTest = function (frameworkName, baseUrl, speedMode, laxMod
 			});
 		});
 
-		/*test.describe('Persistence', function () {
+		test.describe('Persistence', function () {
 			test.it('should persist its data', function () {
 				// set up state
 				page.enterItem(TODO_ITEM_ONE);
@@ -302,13 +302,13 @@ module.exports.todoMVCTest = function (frameworkName, baseUrl, speedMode, laxMod
 				stateTest();
 
 				// navigate away and back again
-				browser.get(otherUrl);
+				browser.get('about:blank');
 				browser.get(baseUrl);
 
 				// repeat the state test
 				stateTest();
 			});
-		});*/
+		});
 
 		test.describe('Routing', function () {
 			test.it('should allow me to display active items', function () {


ColinEberhardt:
Hi @arthurvr - thanks for looking into this. Have you tested your fix across all the ToDoMVC implementations?

    npm run test -- --grep persist

I am still seeing failures:

```
  TodoMVC - ampersand
    Persistence
      √ should persist its data (1417ms)

  TodoMVC - angular-dart
    Persistence
      √ should persist its data (3731ms)

  TodoMVC - angularjs
    Persistence
      √ should persist its data (1483ms)

  TodoMVC - angularjs-perf
    Persistence
      √ should persist its data (1404ms)

  TodoMVC - angularjs_require
    Persistence
      1) should persist its data

  TodoMVC - ariatemplates
    Persistence
      2) should persist its data

  TodoMVC - atmajs
    Persistence
      √ should persist its data (1351ms)

  TodoMVC - aurelia
    Persistence
      3) should persist its data

  TodoMVC - backbone
    Persistence
      √ should persist its data (1354ms)

  TodoMVC - backbone_marionette
    Persistence
      √ should persist its data (1940ms)

  TodoMVC - backbone_require
    Persistence
      4) should persist its data

  TodoMVC - canjs
    Persistence
      √ should persist its data (1470ms)

...
```

However. if I re-run the failed tests individually, they pass!

I never did get to the bottom of this one ;-)

arthurvr:
Should dig in more but I'm not seeing this behaviour.

arthurvr:
> I am still seeing failures:

And yes, that's true, but we should just open issues about that. Those are specific problem or problems with the apps itself (from what I can tell).

samccone:
:+1: I am ok with this change.. 

When (cough https://github.com/tastejs/todomvc/pull/1212) finally lands we will typically be running only one suite at a time which will be nice. 

but I agree, the flakey tests can be resolved there is no question about it. I am all for more coverage in this case :)

ColinEberhardt:
I'd forgot that #1212 was still outstanding - that really needs merging, it will be a great help :-)

PR #1273, opened 2015-04-26

elie29:
If we don't debounce "this.render", the app-view will be rendered 4 times on each todo (visible, change, change:complete and sync). So when we try to check globally all todos, this.render is called a hundred times.


elie29:
@samccone I closed the last PR and I have opened this one, including only the debounce.

Do you want me to commit under another PR, the checkbox problem? Thanks.

samccone:
:+1: thanks for this fix @elie29 and thanks for your work

elie29:
Cheers !

2015-04-29 15:34 GMT+02:00 Sam Saccone <notifications@github.com>:

PR #1274, opened 2015-04-26

elie29:
If we don't debounce "this.render", the app-view will be rendered 4 times on each todo (visible, change, change:complete and sync). So when we try to check globally all todos, this.render is called a hundred times. Once this PR is accepted, I will edit Backbone-requirejs.


samccone:
:+1: thanks again

PR #1288, opened 2015-05-06
test work!

samccone:


tests/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "todomvc-tests",
+  "private": true,
```
arthurvr: There's no deal with including `name` when you make something private.

samccone: :)

arthurvr:
It does seem strange that we're going to have a `tests/` directory and that the actual tests aren't there but in `browser-tests/`.

arthurvr:
I don't really like the `util/` directory name as well. It gives the impression that actual code lives in there. That's not really true, most people will never need these scripts.

sindresorhus:
Maybe:

`browser-tests` => `test`
`util` => `tooling`
`tests` => `tooling`

arthurvr:
Yes. I wonder if we can dump that `travis-runner.sh` in `tooling` then as well?

samccone:
yeah i can drop the util folder and just make a func since it is only used in one spot

samccone:
travis-runner is actually a bad name now.. and can just be test-runner.sh since it will run everything :)

test-runner.sh
```diff
@@ -1,6 +1,14 @@
 #!/bin/bash
 set -o pipefail
 
+
+get_changes ()
+{
+  git remote add current https://github.com/tastejs/todomvc.git && \
```
arthurvr: why does this file use 2 spaces indentation?

samccone:
all updated @arthurvr

arthurvr:
Yiippeeee!

PR #1296, opened 2015-05-13
Updated example for latest WebRx Router changes

oliverw:

PR #1302, opened 2015-05-17
webrx fixes

### LINKED ISSUE #1293: WebRx implementation routing doesn't use hash-bang
The current routing implementation uses the History API which doesn't work for our statically served pages. We need to migrate this to using `location.hash` instead.

samccone:
Fixes #1293
via #1296

### DIFF
--- a/examples/webrx/js/app.js
+++ b/examples/webrx/js/app.js
@@ -123,12 +123,9 @@
 		// setup routing
 		var self = this;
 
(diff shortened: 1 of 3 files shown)


arthurvr:
Thanks Sam. In the original PR we also removed lines 126-129. Wasn't that needed? I :heart: the new CI system by the way. Lovely.

samccone:
fixed :)

arthurvr:
:rocket:

Issue #1234, opened 2015-03-26
Get rid of completed todos count


arthurvr:
As of https://github.com/tastejs/todomvc/commit/a5c2dc6c7c21cd4ef4e5ec7b6da63635f0d34647 the number of completed todos isn't needed anymore in the `clear completed` button. It can be removed from existing apps.

Ref https://github.com/tastejs/todomvc-app-css/pull/7.

arthurvr:
Dropping in the app list so we can easily follow this up. I'm already working on some of them as well as the tests.

- [ ] ampersand
- [ ] angular-dart
- [x] angularjs
- [ ] angularjs_require
- [ ] angularjs-perf
- [ ] ariatemplates
- [ ] atmajs
- [ ] backbone
- [ ] backbone_marionette
- [ ] backbone_require
- [x] batman
- [ ] canjs
- [ ] canjs_require
- [ ] chaplin-brunch
- [ ] closure
- [ ] componentjs
- [ ] cujo
- [ ] derby
- [ ] dijon
- [ ] dojo
- [ ] duel
- [ ] durandal
- [x] emberjs
- [ ] enyo_backbone
- [ ] epitome
- [ ] exoskeleton
- [ ] extjs_deftjs
- [ ] firebase-angular
- [ ] flight
- [ ] gwt
- [ ] jquery
- [ ] kendo
- [ ] knockback
- [ ] knockoutjs
- [ ] knockoutjs_require
- [ ] lavaca_require
- [ ] meteor
- [ ] mithril
- [ ] montage
- [ ] mozart
- [ ] olives
- [ ] plastronjs
- [ ] polymer
- [ ] puremvc
- [ ] ractive
- [ ] rappidjs
- [ ] react
- [ ] react-backbone
- [ ] sammyjs
- [ ] sapui5
- [ ] serenadejs
- [ ] socketstream
- [x] somajs
- [ ] somajs_require
- [ ] spine
- [ ] stapes
- [ ] stapes_require
- [ ] troopjs_require
- [ ] typescript-angular
- [ ] typescript-backbone
- [ ] vanilladart
- [ ] vanillajs
- [ ] vue
- [ ] yui

arthurvr:
Just so I don't forget it, this would also fix https://github.com/tastejs/todomvc/issues/1252.

arthurvr:
[...]

PR #1305, opened 2015-05-20
Update the FOAM example to the new UI, and also to latest FOAM

shepheb:
This splits the code up into four files under a Java-style package hierarchy, as is the modern FOAM convention. 

`build/foam.js` is a concatenated file built using FOAM's build tool, which includes the portions of the framework needed to run TodoMVC, plus the (slightly munged) TodoMVC code itself. I've included the original code (under `js/`) as well as the processed code (`build/foam.js`), and the latter is what executes when the app is loaded.

I know that you don't like junk files included in these examples; if you'd rather I remove the human-readable `js/` versions, I can do that; the files can be read in the FOAM repository either way. The build tool strips unnecessary chunks of the framework, so the total line count is reduced even with the redundant human-friendly files.

Passes lint, and most of the tests. Four are failing due to `StaleElementReferenceError`s; this is because we're sometimes re-rendering the rows but the tests are hanging onto old elements. Is it part of the spec not to re-render, or is this an accident of the tests? Reproducing the tests' intended behavior by hand, they work fine.


shepheb:
Addresses #1234 and #1110.

examples/foam/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "private": true,
+  "dependencies": {
+    "todomvc-common": "^1.0.0",
+    "todomvc-app-css": "^1.0.1"
+  },
+  "devDependencies": {
+    "foam-framework": "0.4.2"
+  }
```
arthurvr: `devDependencies`? That would mean we don't need foam to run the app?

examples/foam/build.sh
```diff
@@ -0,0 +1,15 @@
...
+
+rm -rf $OUT_DIR
+mkdir $OUT_DIR
+node $FOAM_DIR/tools/foam.js foam.build.BuildApp targetPath=$TODO_DIR/build \
+  controller=com.todomvc.Controller \
+  htmlFileName=junk.html \
+  extraModels=foam.ui.BooleanView,foam.ui.DetailView,foam.ui.ActionButton \
+  extraClassPaths=$TODO_DIR/js
+
+rm $OUT_DIR/junk.html
```
arthurvr: is this script really necessary?

examples/foam/package.json
shepheb: Yes, it's true. You need FOAM to run the `build.sh` script which constructs `build/foam.js`. That file contains the app code and the subset of the framework needed to run it.

examples/foam/build.sh
shepheb: Only if we want to be able to rebuild the `build/foam.js` script that is being run. I don't know what the policies are here. It isn't necessary if we just want the code checked in and don't care about rebuilding it.

shepheb:
Still an open question: should I remove the build script? The human-readable code? I don't know what the policy is for this, and I don't know if there are any other frameworks in a similar position I can look to for guidance.

arthurvr:
> Still an open question: should I remove the build script? The human-readable code? I don't know what the policy is for this, and I don't know if there are any other frameworks in a similar position I can look to for guidance.

Haven't dug into FOAM too much but if it's not required to run the app or to develop it later on, the policy would be to remove it. Just like we don't want a grunt/gulp build script in other plugins. If it's totally required, it's okay but try to keep it as simple as possible and make sure it runs as an npm script.

shepheb:
I'll keep the script in the FOAM repo, but remove it from here. I've kept the human-readable code, which I emphasize is **not** required to run the app. Should I remove it too?

PR updated.

samccone:
while I appreciate this footer I think it is not inline with our guidlines :wink: 

screen shot 2015-05-25 at 3 20 57 pm

samccone:
I am getting this console warning, is this expected?

screen shot 2015-05-25 at 3 22 39 pm

examples/foam/js/com/todomvc/TodoFilterView.js
```diff
@@ -0,0 +1,26 @@
...
+				name: 'choices',
+				factory: function() {
+					return [ [ TRUE, 'All' ], [ NOT(this.Todo.COMPLETED), 'Active' ], [ this.Todo.COMPLETED, 'Completed' ] ];
+				}
+			}
+		],
+		methods: [
+			function choiceToHTML(id, choice) {
+				var self = this;
+				this.setClass('selected', function() { return self.text === choice[1]; }, id);
```
arthurvr: `function ()`. Mind running jscs against your app? I see some other code style violations.

arthurvr:
Ping @shepheb

arthurvr:
Hey @shepheb - still interest in finishing this up?

shepheb:
PR updated. jscs is now clean, and I've removed the nonstandard footer.

Sorry for the delay in getting back to this.

examples/foam/index.html
```diff
@@ -2,11 +2,10 @@
...
   <meta charset="utf-8">
   <title>FOAM • TodoMVC</title>
-  <link rel="stylesheet" href="bower_components/todomvc-common/base.css">
-  <link rel="stylesheet" href="bower_components/todomvc-app-css/index.css">
-  <script src="bower_components/foam/build/foam.js"></script>
-  <script src="bower_components/todomvc-common/base.js"></script>
-  <script src="Todo.js"></script>
+  <link rel="stylesheet" href="node_modules/todomvc-common/base.css">
+  <link rel="stylesheet" href="node_modules/todomvc-app-css/index.css">
+  <script language="javascript" src="node_modules/todomvc-common/base.js"></script>
```
arthurvr: Is it a framework specific thing, or what's up with this `language` attribute? It's moot.

sindresorhus: Lol, the `language` attribute was deprecated 15+ years ago.

arthurvr: Yeah, quite lol.

samccone: Looks like it is still in the spec
http://www.w3.org/TR/html4/interact/scripts.html#h-18.2.2

But it looks like since we specify the content type in the header we should not need this.

> Otherwise, if any HTTP headers specify the "Content-Script-Type", the last one in the character stream determines the default scripting language.

shepheb: Yeah, that's an archaism. Even with an uncooperative server that didn't supply the right headers, it should probably be spelled type="application/javascript" now or something. I've removed it.

arthurvr:
Tests don't pass but no new failing specs were introduced so I'm okay with that. Mind taking a look at the console? Looks like there are still some errors.

shepheb:
All console errors are now resolved.

The test failures, for the record, are because the app is re-rendering a few pieces of the DOM that the tests expect to persist. The tests store them in variables, and don't look the elements up afresh in successive tests.

If re-rendering like that is intentionally against the spec, we might be able to work around it. But FOAM does this fairly often with simple controls, recreating them with new details but the same ID.

arthurvr:
:+1: LGTM.

shepheb:
Great! Thanks for working with me on this.

Issue #1285, opened 2015-05-02
Node modules and angularjs tests


blacksonic:
Is there a reason behind committed node_modules folder?
Why is it necessary to maintain a different ```package.json``` for AngularJS tests?

The reason behind these questions is to start increase the quality of the AngularJS example to be more sane and understandable.

arthurvr:
> Is there a reason behind committed node_modules folder?

There is. We use npm to manage front-end dependencies so that makes sense (they should be there on the live site).

> Why is it necessary to maintain a different package.json for AngularJS tests?

I agree in there. We better add the testing dependencies as `devDependencies` in the main angular `package.json`. You in for a PR?

blacksonic:
Im in for a PR, will check it tomorrow.
The site is on GithubPages? Because otherwise a simple ```npm install``` would solve the frontend dependencies.

arthurvr:
> The site is on GithubPages? Because otherwise a simple npm install would solve the frontend dependencies.

A 'simple' npm install in every app folder, yes. We should look into that for the future. We first want to setup some other CI features :) 

A PR would be awesome. Don't forget to update the [`.gitignore`](https://github.com/tastejs/todomvc/blob/master/examples/angularjs/.gitignore).

samccone:
ping @blacksonic

arthurvr:
There we go: https://github.com/tastejs/todomvc/pull/1311 :)

PR #1311, opened 2015-05-26
Angular: update `package.json`

arthurvr:
* Updated todomvc-common to the latest
* Move `angular-route` to `dependencies`. It's needed to run the app so shouldn't be in `devDependencies`.
* Update the test suite to not have a separated `package.json` in its own folder, instead have everything in the main one as dev dependencies. (https://github.com/tastejs/todomvc/issues/1285)

Fix https://github.com/tastejs/todomvc/issues/1285.


addyosmani:
Changes LGTM.

arthurvr:
That was quick. Thanks @addyosmani :)

PR #1314, opened 2015-05-26
Added UI QE POC

Kapslock:
Added UI QE POC


samccone:
Context? We already have a full suite of tests

Issue #1315, opened 2015-05-26
Remove sapui5


samccone:
The app is non-functional currently.  If no one wants to fix it we will remove this example in accordance with https://github.com/tastejs/todomvc/issues/1203

----------------

ping  @schubertha @passy 

any interest in fixing?

##### last (functional) change

104cfde3f5a4b08df3598f4e136a60c80549e4fd

##### Introduced in 

e7fbe334bb09689fc2d14fef348263b3c25cd7b8

arthurvr:
:+1: If nobody comes in to fix this soon let's drop it.

passy:
Oddly, it receives a lot of traffic, but yeah, if no one's around to fix it, let's drop it.

akudev:
Hi,
I'll at least have a look.

akudev:
It appears like the previous code only worked because Array.prototype.push had been overridden by someone and had been returning the resulting array instead of the array length.
Maybe in an external library, which has been removed or fixed recently...

PR #1319, opened 2015-05-28
SAPUI5: fix adding new tasks

akudev:
The return value of Array.push() was expected to be the resulting array,
but it is the length of the array. Not sure how this ever worked, maybe
Array.push() had been overridden wrongly somewhere.

Also update the URL from which SAPUI5 is loaded:
sapui5.hana.ondemand.com is the one to use and is powered by Akamai CDN.

Fixes https://github.com/tastejs/todomvc/issues/1315

### DIFF
--- a/examples/sapui5/index.html
+++ b/examples/sapui5/index.html
@@ -2,9 +2,10 @@
 <html lang="en" data-framework="sapui5">
 	<head>
 		<meta charset="utf-8">
+		<meta http-equiv="X-UA-Compatible" content="IE=edge" />
 		<title>SAPUI5 • TodoMVC</title>
 		<script
-			src="https://sapui5.netweaver.ondemand.com/resources/sap-ui-core.js"
+			src="https://sapui5.hana.ondemand.com/resources/sap-ui-core.js"
 			id="sap-ui-bootstrap"
 			data-sap-ui-libs="sap.ui.commons"
 			data-sap-ui-theme="base">
--- a/examples/sapui5/js/todo/Todo.controller.js
+++ b/examples/sapui5/js/todo/Todo.controller.js
@@ -39,12 +39,13 @@
 				return;
 			}
 
-			this.model.setProperty('/todos/', this.model.getProperty('/todos/')
-				.push({
-					id: jQuery.sap.uid(),
-					done: false,
-					text: todo
-				}));
+			var todos = this.model.getProperty('/todos/');
+			todos.push({
+				id: jQuery.sap.uid(),
+				done: false,
+				text: todo
+			})
+			this.model.setProperty('/todos/', todos);
 
 			this.store.set(this.model.getData());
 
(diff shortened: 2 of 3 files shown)


akudev:
I think this Travis code check failure refers to the underscore in:
`
sap.ui.Device.browser.internet_explorer
`
However, this is an API of the library, so the name of this variable cannot be changed in the TodoMVC app. Should I invert the logic to avoid mentioning "internet_explorer" or can I add a comment making lint ignore this? Or is it fine to merge this code despite the failure?

examples/sapui5/index.html
```diff
@@ -2,9 +2,10 @@
 <html lang="en" data-framework="sapui5">
 	<head>
 		<meta charset="utf-8">
+		<meta http-equiv="X-UA-Compatible" content="IE=edge" />
```
passy: That property is no longer supported. Could you remove it?

samccone:
Hi @akudev thanks so much for fixing this example!

Would you mind adding `//jscs:disable` and then `//jscs:enable` between the lines that are causing the jscs failures.

akudev:
Oh, sorry, I missed the IE=edge comment. Erm... no longer supported? But surely it is still supported in browsers like IE9? I can remove it anyway, no problem.

Is this additional commit (and possibly another one for the IE=edge thing) ok, or should I squash all into one commit?

arthurvr:
> Oh, sorry, I missed the IE=edge comment. Erm... no longer supported? But surely it is still supported in browsers like IE9? I can remove it anyway, no problem.

It doesn't make sense to keep deprecated features in, sorry. I doubt anyone is using todomvc on IE9 anyways :)

> Is this additional commit (and possibly another one for the IE=edge thing) ok, or should I squash all into one commit?

That would be a commit that fixes the previous one, so let's squash.

> Should I invert the logic to avoid mentioning "internet_explorer" or can I add a comment making lint ignore this? Or is it fine to merge this code despite the failure?

I was actually going over some code recently and planning to remove that hack anyways. It isn't true for opera anymore and it's an IE problem, not ours. `jQuery.browser` is deprecated too. We don't deal with that "problem" in any other example.

arthurvr:
> Would you mind adding //jscs:disable and then //jscs:enable between the lines that are causing the jscs failures.

Why? Fix issues instead of ignoring them, no? If we just ignore all code style guide violations, there's no deal with using the linter, right?

samccone:
> I think this Travis code check failure refers to the underscore in:
 sap.ui.Device.browser.internet_explorer 
However, this is an API of the library

no way to change it if it is in the API of the framework :)

arthurvr:
I would actually prefer to just remove the hack (in a separated commit) and just always autofocus. It's not in the app spec and we don't take care of it in any other example. Additionally we shouldn't be suggesting anybody to write code like that (and `jQuery.browser` is deprecated too).

akudev:
Sorry, I messed up my local repo, I did a new pull request at https://github.com/tastejs/todomvc/pull/1320

PR #1320, opened 2015-05-28
SAPUI5: fix adding new tasks, remove browser-dependent autoselect

akudev:
The return value of Array.push() was expected to be the resulting array,
but it is the length of the array. Not sure how this ever worked, maybe
Array.push() had been overridden wrongly somewhere.

The browser detection API conflicted with TodoMVC's code checks and is
anyway undesired.

Also update the URL from which SAPUI5 is loaded:
sapui5.hana.ondemand.com is the one to use and is powered by Akamai CDN.

Fixes https://github.com/tastejs/todomvc/issues/1315

### DIFF
--- a/examples/sapui5/index.html
+++ b/examples/sapui5/index.html
@@ -4,7 +4,7 @@
 		<meta charset="utf-8">
 		<title>SAPUI5 • TodoMVC</title>
 		<script
-			src="https://sapui5.netweaver.ondemand.com/resources/sap-ui-core.js"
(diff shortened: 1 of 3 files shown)


arthurvr:
Looking good to me. If you really feel like getting the example alive again take a look at https://github.com/tastejs/todomvc/issues/1234 and https://github.com/tastejs/todomvc/issues/1110 hint hint :)

akudev:
Yeah, just saw them... :-)

Thanks!

PR #1321, opened 2015-05-29
Fix `frameworkNamePattern` regex

arthurvr:
In the test suite, we had a regex in place to match framework folder names. It didn't include numbers so we couldn't run tests on the `sapui5` app for example.


arthurvr:
If nobody complains I'm going to merge this one. No big deals and I in fact should have committed it directly anyways.

samccone:
ha no worries i was literally just making this change :)

PR #1322, opened 2015-05-30
Update readme.md to add instructions for Visual Studio 2015 RC

svcondor:
As per request I've added instructions for running this example under Visual Studio 2015 RC.

### DIFF
--- a/examples/typescript-angular/readme.md
+++ b/examples/typescript-angular/readme.md
@@ -82,4 +82,52 @@ To compile the TypeScript in this project:
 	# from examples/typescript-angular
 	run `npm run compile`
 
-Or use Visual Studio with the TypeScript plugin.
+<br />
+## Visual Studio 2015 RC
+
+There is no VS template for client-side only apps but the following should get this example running.
+
+>Create a new ASP.NET Web Application solution and choose **empty "ASP.NET 5 Preview Template"**.
+
+>**Uncheck** "Create directory for solution" .   
+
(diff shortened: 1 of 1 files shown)


arthurvr:
Does it make sense to have such a big explanation in that document about this? It seems like not related to comparing MVC solutions anymore.

Please also edit the PR title to be something more useful than `update readme.md`. Using this one we can't even know about what example we're talking about ;)

svcondor:
Sorry, I’m new to GitHub, I’ve changed the PR title.

It would be better if Visual Studio had a suitable template for client-side javaScript apps.   I don’t think they can be more concise without leaving something out.   Would you prefer them in a separate .md file or just as an issue?

R, David.

arthurvr:
Even we add something at all, I'd rather have us linking to an external resource than even having this all in the repo. This isn't helping people comparing MVC frameworks anymore, it's teaching people how to use an IDE.

svcondor:
You’re right it’s not related to comparing MVC frameworks.   I only did it because I thought you’d asked me to.
I’ve only just realized that samccone/todomvc is a fork of tastejs/todomvc.   He asked me to do this but I did the pr on your repo by mistake.

“samccone  commented 9 hours ago
If anyone would like to open a pr and add readme docs around how to run this in visual studio it would be very welcome. Thanks @svcondor”

I’m sorry for all the confusion, I’ll delete the fork if I can work out how.

Regards, David.

PR #1327, opened 2015-06-01
SAPUI5: remove "completed todos" count

akudev:
As per request in https://github.com/tastejs/todomvc/issues/1234


arthurvr:
This looks good. Tests don't pass but not any sapui test passes so I guess it's because of wrong markup (or anything else). Thanks @akudev.

PR #1328, opened 2015-06-01
<pending> Review only

samccone:
Allowing for multiple lookup selectors.


samccone:
This commit should be the first step in allowing for dynamic selectors across the entire app

arthurvr:
@samccone Can we delete this branch?

Issue #254, opened 2012-08-02
Unit testing


sindresorhus:
As discussed in #200 and #134 we are going to require framework submissions to come with unit tests. We'll provide a Mocha testsuite by default, but each framework is free to choose another if that's their best practice.

We already have a good start with the [Mocha tests](https://github.com/addyosmani/todomvc/tree/ea6780dfccf1865c3029ed1f31e4f77fd25637a2/dependency-examples/emberjs_require/js/app/specs) @stas created for us in the Ember+RequireJS app.

Todo:
- Review the above unit tests and check if something is missing
- Make it more generalized and boilerplaty
- Needs documentation on how to implement it in new frameworks
- Figure out which frameworks that should use this boilerplate and which should use something different

FadySamirSadek:
Closed as out of datee

PR #1329, opened 2015-06-01
End 2 end

flightgong:
Wring end 2 end test for the angular project


samccone:
hi @flightgong we already have a test suite for all implementations under `tests` if you are interested in working on them / improving please open a PR into there.

thanks!

arthurvr:
If you'd like to work on app-specific tests, check https://github.com/tastejs/todomvc/issues/254.

PR #1330, opened 2015-06-05
test id => class work

samccone:
this allows for dynamic app discovery and sets some state to use in the other selectors

### DIFF
--- a/tests/page.js
+++ b/tests/page.js
@@ -2,6 +2,7 @@
 
 var webdriver = require('selenium-webdriver');
 
+var idSelectors = false;
 module.exports = function Page(browser) {
 
 	// ----------------- utility methods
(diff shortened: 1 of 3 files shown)


samccone:
kk @arthurvr @sindresorhus this makes the integration tests work across v1 app implementations and v2 100%

:dancers: 

I tested this against my angular2 implementation and everything seems dandy :rose:

tests/page.js
```diff
@@ -37,15 +37,18 @@ module.exports = function Page(browser) {
 	// elements which *might* be present in the DOM, hence the try/get name.
 
 	this.tryGetMainSectionElement = function () {
-		return this.tryFindByXpath('//section[@id="main"]');
+		return this.tryFindByXpath(
+			idSelectors ? '//section[@id="main"]' : '//section[contains(@class, "main")]');
```
passy: Any chance we could turn this into a function instead? Seems like an easy source for errors like this.

samccone:
kk @passy all updated :clap:

tests/page.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 var webdriver = require('selenium-webdriver');
-
+var idSelectors = true;
```
arthurvr: It would be more logical if this defaults to `false`. The usage of IDs is temporary, all examples will need to be switched over one day.

arthurvr: ping @samccone - don't you think so?

samccone:
updated :dog: 

screen shot 2015-06-17 at 10 07 02 am
screen shot 2015-06-17 at 10 06 52 am

arthurvr:
lgtm

samccone:
:kiss:

passy:
:sparkles:

PR #1340, opened 2015-06-14
Updates to the `react-backbone` example

arthurvr:
Changes include:

* update todomvc-common
* update react to the latest version and fix all deprecations
* move to using the `classnames` module directly instead of the build-in version. (this is [encouraged](https://facebook.github.io/react/docs/class-name-manipulation.html))
* move away from using the `react-with-addons` file as we don't use any addons

cc @vasanthk @petehunt looking good?


vasanthk:
@arthurvr Looks good to me. :+1:

arthurvr:
Thanks. Merging. These changes are rather trivial anyways.

Issue #1069, opened 2014-11-19
Update Wanted: Ember


passy:
The app is on Ember 1.6.1, the latest is 1.9.0-beta.3. Additionally, we're on ember-data#1.0.0-beta.7 (latest is 1.0.0-beta.11). I believe the latter has some breaking changes.

Anyone up for this? :)

nikolas:
I agree @phillipsj, ember-cli should be a different application in TodoMVC. It's still valuable to have the existing ember.js application up because there's still use cases for people wanting to set an Ember app up the more traditional, manual way.

For example, I'm making a Django application and I want the Ember application to be defined in a Django template. In that case, it seems to make more sense to not use ember-cli. I like using ember-cli when it's appropriate, though.

arthurvr:
Regarding ember CLI, I personally think it's out of scope for TodoMVC. Ember CLI isn't about an MV* framework anymore. It's about the build script and developer tooling used for building apps, with a specific framework.

So, I'd personally say we should just update to the latest ember but an Ember CLI app is out of scope. Though, maybe somebody should just kick off an app in a separated repo. What does the rest think?

iheanyi:
So we're in agreement that there needs to be a completely separate ember-cli example, then? Really interested in this, so just wanna make sure. :)

samccone:
yep! @iheanyi I think a great initial thing to start with would be

https://github.com/ember-cli/ember-cli-todos

open a a PR and we can go from there.. thanks a million :revolving_hearts:

samccone:
DONE!

PR #1343, opened 2015-06-16
Update emberjs example versions

spastorino:
This helps avoiding an infinite PUTs loop when used with the RESTAdapter.

The issue seems to be the same to the one described here ...
http://discuss.emberjs.com/t/save-post-update-put-ember-data-infinite-loop-when-upgrading-to-ember-1-0-0-4-and-ember-data-1-0-0-beta-3/2824

Hat tip @san650

Just in case you want to check the resources I've used ...

http://builds.emberjs.com/release/ember.js
http://builds.emberjs.com/beta/ember-data.js
http://builds.emberjs.com/release/ember-template-compiler.js
http://code.jquery.com/jquery-2.1.4.js

### DIFF
--- a/examples/emberjs/node_modules/ember-data/ember-data.js
+++ b/examples/emberjs/node_modules/ember-data/ember-data.js
@@ -1,20 +1,7 @@
 (function() {
     "use strict";
-    /**
-      @module ember-data
-    */
-
-    var ember$data$lib$system$adapter$$get = Ember.get;
-
-    var ember$data$lib$system$adapter$$errorProps = [
-      'description',
-      'fileName',
-      'lineNumber',
-      'message',
-      'name',
-      'number',
-      'stack'
-    ];
+    var ember$data$lib$system$model$errors$invalid$$create = Ember.create;
+    var ember$data$lib$system$model$errors$invalid$$EmberError = Ember.Error;
 
     /**
       A `DS.InvalidError` is used by an adapter to signal the external API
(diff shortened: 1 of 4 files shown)


passy:
Are those just patch release updates to what's specified in the `package.json`, ie. what you get from running `npm update`?

spastorino:
@passy yeah I would do npm update, the thing is the issue is with ember-data, we would need to override that one. Do you want me to run npm update and override ember-data.js?

spastorino:
I've updated the PR reflecting exactly that. Now it's the result of npm update + curl http://builds.emberjs.com/beta/ember-data.js > node_modules/ember-data/ember-data.js

passy:
I'd like to get input from the others, but I'm slightly hesitant to including dependencies that aren't published on npm yet.

arthurvr:
> I'd like to get input from the others, but I'm slightly hesitant to including dependencies that aren't published on npm yet.

Yep, we should try to push them to publishing it instead.

spastorino:
I mainly updated ember-data because it's there anyway, and the one there has bugs.
But as you guys wish.

spastorino:
A bit more context ember-data was added here https://github.com/tastejs/todomvc/commit/26d239c2cd7e7333dfeafcceac1306f80036f412 before it was being installed using bower.

/cc @tomdale @wycats

tomdale:
We have plans to distribute Ember Data on npm but for now the official distribution channel is Bower. I'll reiterate that I think the best way to give people a taste of Ember is to use Ember CLI to build the app (https://github.com/tastejs/todomvc/issues/1069#issuecomment-63800583); the whole point of Ember is that it's a complete front-end stack and abstracts away painful dependency management stuff like this.

wycats:
I agree with @tomdale

passy:
I'd be in favor of migrating the example to ember-cli if it represents the
best practices for Ember development. Comparability isn't really meaningful
if the app would not actually be built that way.

What do the others think?

addyosmani:
I agree with @passy. If the idiomatic way to build all Ember apps is using ember-cli, then I'm in favour of us migrating the app in that direction.

tomdale:
@addyosmani That's awesome, do you want me to try to rustle up some community members to help out with this? I'm not sure where to get started.

addyosmani:
@tomdale That would be incredibly helpful. One option would be to do a fresh implementation following our spec using ember-cli, using the current implementation of the Ember app for inspiration as needed. We can help with reviews/guidance for spec compliance. Love to see this through!

arthurvr:
Over at https://github.com/tastejs/todomvc/issues/1069 some community members already showed their interest in building something up too.

spastorino:
What about this repo https://github.com/ember-cli/ember-cli-todos and/or asking these people for help? /cc @tomdale @wycats

arthurvr:
There was an Ember CLI example submitted: #1360. Let's close this PR in favor of that one.

PR #1356, opened 2015-06-28
🎅 Update angular to latest

samccone:
The motivation here is partially to be able to confirm a memory leak in Angular core that I would like to bubble up the chain.

### DIFF
--- a/examples/angularjs/js/controllers/todoCtrl.js
+++ b/examples/angularjs/js/controllers/todoCtrl.js
@@ -23,10 +23,9 @@ angular.module('todomvc')
 		// Monitor the current route for changes and adjust the filter accordingly.
 		$scope.$on('$routeChangeSuccess', function () {
 			var status = $scope.status = $routeParams.status || '';
-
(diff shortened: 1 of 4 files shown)


examples/angularjs/js/controllers/todoCtrl.js
```diff
@@ -23,10 +23,9 @@ angular.module('todomvc')
 		// Monitor the current route for changes and adjust the filter accordingly.
 		$scope.$on('$routeChangeSuccess', function () {
 			var status = $scope.status = $routeParams.status || '';
-
```
samccone: automated whitespace trim... I can revert or split this if you guys want

passy:
lgtm :+1:

examples/angularjs/js/controllers/todoCtrl.js
arthurvr: This isn't trimming whitespace, it's removing a blank line which was there for readability so I'd prefer it to stay.

arthurvr:
lgtm

PR #1367, opened 2015-07-07
🐛 Fix test routing selector bug

samccone:
should have caught this earlier, this should fix your issues @arthurvr 

:clap:


arthurvr:
:+1:  Thanks!

PR #1369, opened 2015-07-08
Drop the YUI app

arthurvr:
From the official site:

>  The YUI library is no longer actively maintained.

We shouldn't suggest anybody to use a framework which isn't active anymore.

Ref http://yahooeng.tumblr.com/post/96098168666/important-announcement-regarding-yui


samccone:
:+1: lol the test failure

sindresorhus:
:+1:

PR #1363, opened 2015-07-02
Update `react` example to use classes

arthurvr:
Updated the react example to the latest `todomvc-app-css`. This includes the change to use classes instead of IDs. 

@samccone Seems like all routing tests fail. Looks like a problem with the testsuite. Can you take a look?

PR #1370, opened 2015-07-08
fix stale element errors

samccone:
By converting the operations to be a single op vs iterating over a found set of nodes and touching the DOM this removes the issue of stale elements. 

----

xref #1363 
https://travis-ci.org/tastejs/todomvc/builds/70054158#L739


arthurvr:
lgtm :+1: Neat!

PR #1372, opened 2015-07-11
update `vue` example to use classes

arthurvr:


samccone:
Yep, I would just do a quick test on my cI  branch with this before merging

arthurvr:
> I would just do a quick test on my cI branch with this before merging

Just tested and it seems to works fine. Yay!

PR #1375, opened 2015-07-12
Vanilla dart app updates

arthurvr:
* Use npm to manage dependencies, adopt the new UI (#1110)
* Don't suggest to use the deprecated dart editor.
* Fix tests to point to the right directory.
* Update to classes (`todomvc-app-css@2.0`)


samccone:
@arthurvr so, can you add a .jscsignore and see if you can get this passing locally, when you do an you post the command you ran here so that we can update the linter to respect this file?

arthurvr:
Everything passing now. FYI, there's no such thing as a [`.jscsignore`](https://github.com/jscs-dev/node-jscs/issues/73) file.

samccone:
kk :+1:

samccone:
CI makes this so much easier now :mask: :smile:

robbbz:
Yes thanks for doing that, Sam.

PR #1382, opened 2015-07-16
Ractive.js: Fix documentation links

darthmaim:
The 60 second setup and the docs have been moved from the wiki to the website.

### DIFF
--- a/examples/ractive/readme.md
+++ b/examples/ractive/readme.md
@@ -7,9 +7,9 @@
 
 ## Learning Ractive.js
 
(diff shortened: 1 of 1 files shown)


arthurvr:
Thanks!

samccone:
@arthurvr Looks like some failing specs on ractive we should file these to be addressed

robbbz:
@samccone Keep unrelated discussions out of this ticket, please.

samccone:
@robbbz It is not unrelated but thanks for the input

samccone:
To clarify the doc fix exposed the fact that this implementation fails on CI so I used it as a call to action to open new issues documenting this failure.

arthurvr:
@robbbz Can you open one? Thanks.

PR #1396, opened 2015-08-04
fix: Finish renaming stopEditing to saveEditing

andreaswachowski:
(follow-up to 233db94c1333eda411c70076226df6be76635edc)


arthurvr:
@passy Can you review this?

passy:
@andreaswachowski Thanks, good catch!

passy:
For the record: The failing CI is accurate. There's a delay between before changes are persisted and if you hit refresh quickly after making a change, they're lost.

PR #1399, opened 2015-08-09
Angular: Use $resource instead of $http in api client

kaidokert:
As far as i understand, $resource usage is encouraged whenever the backend api fits it reasonably well.

I wrote up $resource usage in as minimal line count change as i could.
In the process:
- updated angular, angular-resource and angular-router to latest stable
- found that there is redundant code in $http handling returning never used values, which i cut
- learned a bit about error handling and fixed a small bug with put() which was doing undo when it shouldnt, as its handled by controller
- and realized that to make best use of $resource, controller should be simplified as well. the back and forth angular.copy() should not really be needed.
- made a small node.js-only server ( no express, just plain node apis ) to test. Can PR separately if this is useful, not sure where the original server implementation is ? 
- found that `delete all completed` has special meaning in how the API call is made with no parameters against `DELETE api/todos`, not sure if this is intentional

### DIFF
--- a/examples/angularjs/index.html
+++ b/examples/angularjs/index.html
@@ -66,6 +66,7 @@
 		<script src="node_modules/todomvc-common/base.js"></script>
 		<script src="node_modules/angular/angular.js"></script>
 		<script src="node_modules/angular-route/angular-route.js"></script>
+		<script src="node_modules/angular-resource/angular-resource.js"></script>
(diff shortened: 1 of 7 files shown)


samccone:
Thanks @kaidokert I have reviewed, squashed the redundant commits and merged. 

Great work :clap:

PR #1400, opened 2015-08-09
Improve tests for asynchronous applications

smasala:
Not a clean solution, as the original comment in the test points out. However, it helps ensure that more asynchronous applications pass the unit tests


arthurvr:
Thanks for your PR. Except from firebrick does any other app run into this problem?

arthurvr:
I don't think any other examples than FireBrick ran into this problem, and since we decided not to have a FireBrick example I'm going to close this. Thanks though.

smasala:
Hi @arthurvr, thanks for your comments. Yes the KnockoutJS example ran into this, that how I realised it was a timing issue in the test and not my framework. Firebrick also uses Knockout for its data binding. Without the changes in this PR, Knockout JS also fails when attempting the persistent data test.

PR #1403, opened 2015-08-11
Remove helper and view files from tests

kentcdodds:
I'm in the process of rewriting/tooling this in ES6 for a workshop I'm doing and I realized that you don't actually need either of these files for the tests to run 👍

PR #1404, opened 2015-08-11
Fix jscs issues found in #1403...

kentcdodds:
I am so sorry that this results in a bunch of whitespace change and git history nuking 💣

### DIFF
--- a/examples/vanillajs/js/controller.js
+++ b/examples/vanillajs/js/controller.js
@@ -9,40 +9,40 @@
 	 * @param {object} view The view instance
 	 */
 	function Controller(model, view) {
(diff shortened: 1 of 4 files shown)


kentcdodds:
Once this gets merged, I'll rebase it into #1403 and force push. Hopefully then it will get in. Thanks!

samccone:
oh wow looks like this does not even pass specs either! but thanks for this!!!

kentcdodds:
Thanks! I'll go ahead and rebase the other one and update the PR. Hopefully we can figure out why the tests are failing later.

samccone:
:+1:

PR #1406, opened 2015-08-11
Fix vanillajs spec compliance error

samccone:
found in: #1403


samccone:
schweet :goat: :rabbit:

PR #1412, opened 2015-08-14
updated OpenUI5 sample app

agraebe:
changed name to official OpenUI5 instead of SAPUI5 and updated to
latest library version.


agraebe:
squashing my commits before I'll go ahead..

PR #1423, opened 2015-08-24
Changing redraw strategy while handling input events

marcolamberto:
Changing redraw strategy to "none" while handling input events in order to avoid focus issues (caret forced to the input value end).

### DIFF
--- a/examples/mithril/js/controllers/todo.js
+++ b/examples/mithril/js/controllers/todo.js
@@ -8,7 +8,7 @@ app.controller = function () {
 	this.list = app.storage.get();
 
 	// Update with props
-	this.list = this.list.map(function(item) {
+	this.list = this.list.map(function (item) {
 		return new app.Todo(item);
 	});
 
@@ -53,6 +53,10 @@ app.controller = function () {
 	};
 
 	this.doneEditing = function (todo, index) {
+		if (!todo.editing()) {
+			return;
+		}
+
 		todo.editing(false);
 		todo.title(todo.title().trim());
 		if (!todo.title()) {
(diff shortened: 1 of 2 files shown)


samccone:
@marcolamberto mind squashing this into a single commit?

Also can you describe the steps to reproduce the bug that this is fixing?
Thanks!

marcolamberto:
Ok, I'll squash them. ;)

I've just changed how the view is rendered. In the current "production" code when the input field is displayed the view is re-rendered automatically every time an "oninput" or "onkeyup" event handler is triggered. This is the default behaviour. Btw redrawing the view during a "keyup" event loose the focus and the caret position. The current code seems trying to fix this behaviour by forcing the focus again and moving the selection to the end. The user experience is horrible, just try changing something in the middle of the text.

arthurvr:
Thanks @marcolamberto! That seems like a valid concern. @taylorhakes can you maybe review this? I don't think any of us has many experience with mithril. 

> @marcolamberto mind squashing this into a single commit?

Not a big deal but those commits totally made sense standalone. Now we'll end up with code style changes in random commits.

marcolamberto:
@arthurvr I agree, I usually like having small and specific commits, possibly with a green build. It's not a big deal this time, I've just added a couple of spaces. The other two commits could be considered a single one. :D

samccone:
Sorry my mistake I also like commits to be split by intent. By baddddd 😭

marcolamberto:
@samccone no worries! :D

taylorhakes:
@samccone @marcolamberto This looks good to me. I tested it locally and the issue is solved. Thanks for fixing it!

marcolamberto:
@taylorhakes you're welcome. ;).

arthurvr:
Awesome! Thanks for the fix, @marcolamberto! And thanks for reviewing @taylorhakes!

PR #1425, opened 2015-08-25
Update readme to include a note about unminified libs

### LINKED ISSUE #378: Unminified vendor libs?
Our current policy is to have vendor libs minified for performance and size reasons:

> Use minified version of third-party code and put it in the lib folder.

Today, most projects have a build system that concats and minifies the code anyway, and having unminified code, even for vendors libs, helps a lot when debugging.

I would like to change our policy to require everything to be unminified, rather than minified.

@addyosmani  What do you think?

arthurvr:
Fix #378


addyosmani:
LGTM

Issue #1257, opened 2015-04-07
Polymer items array are not in sync with (un)checked item


gnjec:
Clear completed button shows incorrect value.

This issue can be resolved like this:
```javascript
    // td-item.html
    ...
    itemChangeAction: function() {
	    setTimeout(function(self){self.fire('td-item-changed')}, 0, this);
	    // this.fire('td-item-changed');
    },
```
but there must be a better way :)

gnjec:
This is in Firefox. 
In Chrome and IE11 is ok.

arthurvr:
Thanks for reporting this! I'm able to reproduce it. The count inside of the clear completed button itself is going to get removed (https://github.com/tastejs/todomvc/pull/1245), but this is still worth looking into. The button appears at the wrong moments it seems. @addyosmani you've got an idea?

gnjec:
![1257](https://cloud.githubusercontent.com/assets/11833945/7067524/28b95348-decc-11e4-8e5f-bbca59a7cad9.png)

addyosmani:
Confirmed this is an issue. @arthurvr I would say we should defer any Polymer 0.5.x related fixes for a few weeks and just switch to Polymer 0.8/0.9. The syntax has changed sufficiently that it's not worth us patching this up. We're working on a new TodoMVC implem and we'll PR a refresh when it's ready.

arthurvr:
Awesome. Thanks Addy.

samccone:
Pretty sure my pr resolved this @arthurvr 😎

arthurvr:
@samccone's PR landed so going to close this. Thanks again for the report.

Issue #1342, opened 2015-06-16
Polymer 1.0


MunchyYDL:
Hi guys,

I was interested in converting the Polymer example to support the 1.0 release of the library.

Is there a need to keep the 0.5 version as a separate example to show the differences in syntax between the releases? Or do you only want the latest to be present? I couldn't find anything in the guidelines concerning this.

Happy to be able to contribute with this if you want it. :)

arthurvr:
Somebody at the Polymer team is working on it so no need to duplicate work. Going to keep the ticket open till that's finished.

> Is there a need to keep the 0.5 version as a separate example to show the differences in syntax between the releases? Or do you only want the latest to be present?

Only latest :)

arthurvr:
@addyosmani We wanted this in 1.4 so maybe you can give us a quick status update? If the team can't work on it we can still do it in-house. Thanks :wink:

sindresorhus:
ping @addyosmani

samccone:
I can do this @addyosmani

arthurvr:
The PR landed so closing this now. Thanks everyone.

samccone:
it was reverted due to deploy issues, reopening

samccone:
whoops nm

PR #1426, opened 2015-08-25
🍕🌟Polymer 1.0 🌟🍕

samccone:
welp here it is the 1.0 move, there are quite a few things in this upgrade that I am not too happy with however I will ping some people for some input.

-------

I am not going to the the ~~UI or~~ ID migration in this PR that will be after

------

~~CI Will be green soonish. The initial implementation was never compliment so I'm working on it~~

CI is :green_apple: ~~on every planet except for CI 😭 investigating~~


arthurvr:
Ref gh-1342

arthurvr:
Does this also fix https://github.com/tastejs/todomvc/issues/1257?

samccone:
friendly ping @addyosmani :wink:

addyosmani:
Reviewing today!

addyosmani:
There are a few places here where you're qSA'ing against elements inside a `dom-module` rather than using the `this.$.<ID of thing>`.  I understand doing this for nested elements but where it's not deeply nested, is there a reason for that preference? :lollipop:

samccone:
re:
> qSA'ing against elements inside a dom-module

 oh so when an element is within a `dom-if` it looks like the this.$. never picks up / caches the element :( That is why I am doing it that way :smile:

addyosmani:
Changes here now LGTM. Nice work, @samccone! :cake: :cupid:

samccone:
:+1: :tada: :boom:

Issue #812, opened 2014-02-02
Remove routing mark


sindresorhus:
This one is also not very useful. We should instead strive to get all the apps to implement routing.

Blocked by `Let's also check if we want to keep all of the current examples.` in https://github.com/tastejs/todomvc/issues/735

arthurvr:
> Agreed - the majority do implement routing, so I would push to have all implement it.

Shall I ping the authors of all apps that don't have routing yet in this issue? It's getting time to implement it :)

sindresorhus:
Yes.

addyosmani:
Yes please!

arthurvr:
Hey folks! Could y'all checkout this issue? We'd love to have routing integrated in all apps. You can use the existing apps as example. Be sure to read the [app spec](https://github.com/tastejs/todomvc/blob/master/app-spec.md#routing) as well.

* [ ] angular-dart: @passy 
* [ ] typescript-backbone: I was unable to figure out who built this one. Was it @addyosmani?
* [ ] cujo: @briancavalier
* [ ] olives: @podefr
* [ ] duel: @mckamey
* [ ] dijon: @creynders

Thanks!

podefr:
hey, thanks for the ping. I'll add routing to Olives and will create a pull request.

arthurvr:
Great, thanks!

mckamey:
@arthurvr I am interested in updating for DUEL. Life has been busy but will check it out soon.

arthurvr:
Hey @podefr, is there still interest in picking this up? Thanks!

podefr:
Yes @arthurvr, I'll take care of this

arthurvr:
ping @podefr :)

podefr:
@arthurvr I can't give you an ETA, when do you need this for? It may take a couple of months. The problem isn't the router as I've got this working somewhere in a new branch, but I need to implement a new feature to be able to filter a model without breaking data-binding, and this may take a while due to priorities.

arthurvr:
@podefr No problem I guess :) Keep us updated.

arthurvr:
@mckamey still interested?

mckamey:
@arthurvr, yes still interested. I have a newborn at home so life is quite busy.

arthurvr:
No worries @mckamey. Congrats!

arthurvr:
ping @mckamey

mckamey:
[...]

PR #1427, opened 2015-08-30
DUEL: App Update (#1110), cancel editing on escape keypress (#789), and routing support (#812)

mckamey:
Also updates TodoMVC to latest style/structure.


examples/duel/.gitignore
```diff
@@ -1 +1,3 @@
 target/
+node_modules/
+www/cdn/debug/
```
arthurvr: missing a newline

examples/duel/.gitignore
```diff
@@ -1 +1,3 @@
 target/
+node_modules/
```
arthurvr: You'll have to build up a proper gitignore (*[example](https://github.com/tastejs/todomvc/blob/master/examples/dojo/.gitignore)*) and commit the dependencies needed to run the app.

examples/duel/package.json
```diff
@@ -0,0 +1,10 @@
+{
+	"private": true,
```
arthurvr: Wrong indentation, this looks like 1 space.

examples/duel/src/main/webapp/css/base.css
```diff
@@ -0,0 +1,141 @@
...
+@media (min-width: 899px) {
+	.learn-bar {
+		width: auto;
+		padding-left: 300px;
+	}
+
+	.learn-bar > .learn {
+		left: 8px;
+	}
+}
```
arthurvr: Does this styling really need to live in this folder? Why not just let it be in node_modules? Or are modifications to the styles required?

examples/duel/src/main/webapp/js/todos/controller.js
```diff
@@ -85,20 +57,28 @@ var todos = todos || {};
...
 	// event handlers
 	todos.actions = {
-		addBlur: function () {
-			add(this);
-		},
-
 		add_keypress: function (e) {
-			if (e.keyCode === ENTER_KEY) {
-				add(this);
+			switch (e.keyCode) {
```
arthurvr: Why change this from an `if` statement to a `switch`? Seems still easier to simply do it as before, no?

examples/duel/www/index.html
```diff
@@ -1,17 +1,18 @@
...
+	<footer class="info">
 		<p>Double-click to edit a todo</p>
 		<p>Ported to <a href="http://duelengine.org">DUEL</a> by <a href="http://mck.me">Stephen McKamey</a></p>
 		<p>Part of <a href="http://todomvc.com">TodoMVC</a></p>
 	</footer>
-	<script src="./cdn/d10574a720ccdf2d2c810f0c45afb5f159264a1f.js"></script>
+	<script src="./cdn/916ee1b55ac72cc1cc01404ca8e13a695228cc13.js"></script>
 
 </body>
 </html>
```
arthurvr: missing a newline

examples/duel/www/index.html
```diff
@@ -1,17 +1,18 @@
 <!DOCTYPE html>
```
arthurvr: On the while, can you change this to be lowercase? `<!doctype html>`

arthurvr:
Thanks for the PR! Neat work.

Would've been better to do https://github.com/tastejs/todomvc/issues/1110, #789, the dependency update and the update to classes in different commits. This way it's difficult to review and kinda noisy ;)

In another commit, can you please add the compiled files to the JSCS ignores list? The CI fails on JSCS atm.

mckamey:
Ah #1110. I looked for when that was done and couldn't find it so I just included them together. Yes I should have done separately.

examples/duel/.gitignore
mckamey: What is "improper" about this one? It excludes the dirs that I don't want included.

The dependencies needed to build the app are included in the webapp tree. See how package.json copies them after `npm install`.

examples/duel/src/main/webapp/css/base.css
mckamey: No mods but the DUEL build system cannot reach above the root. package.json puts them here as part of `npm install`. Similar to the angular2 example.

examples/duel/src/main/webapp/js/todos/controller.js
mckamey: I personally think that for `keyCode` blocks a switch makes more sense. By the time you get a few then it becomes easier to just read the case labels than walking the if-else chain.

Arguably only two can go either way. Easy to switch back.

examples/duel/www/index.html
mckamey: That's a generated file. Would require a change to DUEL. Funny I'd forgotten that it does caps. I may make that change in DUEL but not this commit.

examples/duel/www/index.html
mckamey: Again index.html is a generated file and `examples/duel/src/main/resoruces/HomePage.duel` is the source.

samccone:
awesome work @mckamey! Looking forward to this one landing. Thanks again

mckamey:
I'm not sure what/where JSCS ignores are. Any pointers?

samccone:
> EDIT: do I just add them here? /.jscsrc

yep :)

mckamey:
How do I run the linter without having to commit. Looks like some minor tweaks in there I need to fix.

samccone:
@mckamey you can from the tooling directory run

`jscs "-c" "../.jscsrc" "../examples/duel"`

mckamey:
@samccone I just get `jscs: command not found`

examples/duel/.gitignore
arthurvr: Oh, alright. The other apps just pull use the deps out of the node_modules folder so that's why I asked. :)

examples/duel/www/index.html
arthurvr: Oh, alright :+1:

samccone:
hey @mckamey looks like we are really close to getting this to pass CI :sparkles: 
The integration tests are currently failing on the last routing tests.

The tests expect a html structure

```html
<ul class="filters">
				<li>
					<a class="selected" href="#/">All</a>
				</li>
				<li>
					<a href="#/active">Active</a>
				</li>
				<li>
					<a href="#/completed">Completed</a>
				</li>
			</ul>
```

(don't worry if you use IDs here it will still work)

Does this make sense?

Thanks again!!

mckamey:
I was holding off on doing the routing for a different PR for #812 but it was easier than I thought. Hoping that fixes the CI tests too.

samccone:
no biggie @mckamey it looks like some flake... not a big deal :) we can address it later.. thanks so much for all this work!

examples/duel/www/index.html
mckamey: This is now generated as lowercase.

samccone:
thanks @mckamey :ship: ing this one

PR #1430, opened 2015-09-03
Replace deprecated Controller

dsumac:
Marionette has marked `Controller` object as a deprecated feature
Instead of using `Controller`, Marionette encourages you to specify your callbacks on a plain 
Javascript object:
http://marionettejs.com/docs/v2.4.2/marionette.controller.html

I propose to use Marionette Object instead of a plain js object. 
Object incorporates many backbone conventions and utilities like initialize and Backbone.Events.
http://marionettejs.com/docs/v2.4.1/marionette.object.html


passy:
LGTM :+1:

samccone:
👍 thx

PR #1431, opened 2015-09-04
Update marionette version to 2.4.2

dsumac:
Just update

### DIFF
--- a/examples/backbone_marionette/node_modules/backbone.marionette/lib/backbone.marionette.js
+++ b/examples/backbone_marionette/node_modules/backbone.marionette/lib/backbone.marionette.js
@@ -1,6 +1,6 @@
 // MarionetteJS (Backbone.Marionette)
 // ----------------------------------
-// v2.4.1
+// v2.4.2
 //
 // Copyright (c)2015 Derick Bailey, Muted Solutions, LLC.
 // Distributed under MIT license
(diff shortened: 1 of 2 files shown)


arthurvr:
Why? We use a caret so the dependency bump isn't needed. Are we running into any concrete issues?

addyosmani:
Same comment as Arthur. I would have thought we would pick up the correct version via caret here.

dsumac:
I have seen this commit 
https://github.com/tastejs/todomvc/commit/9d98ba63bd4a566b42d6a9e1cfaa0528c760b2c1 and I have thought that the strategy was bump the last version. It is a lack of understanding ?
(So, we aren't into an concrete issue.)

dsumac:
About versions, why define backbone dependency (and underscore) in the package.json ?
It's dependencies are defined in the marionette's package.json. Is there a particular purpose to make that ?

arthurvr:
> I have seen this commit 
> 9d98ba6 and I have thought that the strategy was bump the last version. It is a lack of understanding ?

That commit was part of updating some other stuff in the marionette app (https://github.com/tastejs/todomvc/pull/1185). If there are no issues or code changes required there's no point in simply updating dependencies for a patch bump. Doing so would be useless and way too noisy. Thanks for the suggestion though.

> About versions, why define backbone dependency (and underscore) in the package.json ?
> It's dependencies are defined in the marionette's package.json. Is there a particular purpose to make that ?

Because our app explicitly uses those (using `<script>` tags). Besides from that you can't depend on the location of sub-dependencies as npm changed the behaviour of those multiple times in recent versions.

dsumac:
Thanks for the feedback :thumbsup:

PR #1437, opened 2015-09-12
Remove deprecated code and concepts from Backbone.Marionette example

### LINKED ISSUE #1436: Marionette example uses deprecated features
The Backbone.Marionette example is currently making heavy use of modules, as well as using the wreqr application channel.  Both of these features have been deprecated by the Marionette team.

The recommended replacement for modules is to use a module loader (webpack + es6 seems to be popular with the Mn community, but could also be RequireJS/CommonJS etc).  For the app channel, the recommendation is to use the Wreqr or Backbone.Radio API independently of the application object.  Backbone.Radio is considered the best practice, so that would probably be the correct thing to include.

I have a working fork of the current example that I'd be happy to clean up and submit as a PR, but wanted to clarify one thing first.

Is using ES6 modules + webpack (could be precompiled so that users wouldn't need a compile step, and build task could be done through npm if they wanted to hack on it) to high a bar?  I notice that several of the frameworks that have that as a best practice (specifically Ember and React) are not using modules, while others (Aurelia, Angular2) are.  Would a PR for Mn based on webpack be acceptable?

If not, would commonJS be better, or is the preference for globals based solutions only?

benmccormick:
This commit would close #1436, and remove deprecated features from the Marionette example.  The in depth explanation is bulleted below, but the basic gist is that Modules and Wreqr are removed, code is now organized using a global namespace, with Backbone.Radio for messaging, and I cleaned up the file structure as part of the move to a global namespace.

This contains the following: 
- Removed modules, replacing them with a common global namespace
- Added Backbone.Radio as a new dependency
- Removed usage of Wreqr, replacing it with Radio
- Renamed TodoList.js to Router.js to more accurately reflect its
  function
- Moved all application initialization into TodoMVC.Application
- Moved FilterState code into its own file
- Removed inline script in index.html and moved inititialization code
  into TodoMVC.js

### DIFF
--- a/examples/backbone_marionette/js/TodoMVC.Application.js
+++ b/examples/backbone_marionette/js/TodoMVC.Application.js
@@ -1,0 +1,27 @@
+/*global Backbone, TodoMVC:true */
+
+var TodoMVC = TodoMVC || {};
+
+(function () {
+	'use strict';
+
+	var TodoApp = Backbone.Marionette.Application.extend({
+		setRootLayout: function () {
+			this.root = new TodoMVC.RootLayout();
+		}
+	});
+
+	// The Application Object is responsible for kicking off
+	// a Marionette application when its start function is called
+	//
+	// This application has a singler root Layout that is attached
+	// before it is started.  Other system components can listen
+	// for the application start event, and perform initialization
+	// on that event
+	TodoMVC.App = new TodoApp();
+
+	TodoMVC.App.on('before:start', function () {
+		TodoMVC.App.setRootLayout();
+	});
+
+})();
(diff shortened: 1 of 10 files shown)


examples/backbone_marionette/js/TodoMVC.Application.js
```diff
@@ -0,0 +1,27 @@
+/*global Backbone, TodoMVC:true */
+
+var TodoMVC = TodoMVC || {};
+
+(function () {
+	'use strict';
+
+	var TodoApp = Backbone.Marionette.Application.extend({
```
samccone: heh we *could* just use an object instead of an app since app is really pointless currently ;)

samccone: (we are only using start but triggerMethod on Object would work just as well right¿)

paulfalgout: `Application` also currently helps setup the Inspector

samccone: Fair enough 😊

samccone:
lgtm thanks a bunch @benmccormick

PR #1441, opened 2015-09-19
Add a new framework: GopherJS + Humble

albrow:
### About

[GopherJS](https://github.com/gopherjs/gopherjs) compiles [go](http://golang.org/) to javascript code which can run in the browser. [Humble](https://github.com/go-humble/humble) is a collection of loosely-coupled tools written in go designed to build client-side and hybrid web applications.

There is a [Live Demo](http://d3cqowlbjfdjrm.cloudfront.net/) available for anyone to try.

### Why should this implementation be included?

It's unique! All of the source code is written in pure go. As far as I know, [no other GopherJS app has been submitted so far](https://github.com/tastejs/todomvc/pulls?utf8=%E2%9C%93&q=is%3Apr+gopherjs+). The GopherJS & Humble implementation brings something to the table that most people have not seen before, and many people do not even know is possible. If the goal of the TodoMVC project is to help developers discover what frontend frameworks exist in the world, I believe this example will make a great addition.

Both GopherJS and Humble are actively maintained, well-documented, and well-tested. The application itself is full of source code comments, and the README helps you get the app up and running.

### Tests

All the tests pass. Here is the output of `npm run test -- --framework=gopherjs-humble`:

```
> todomvc-browser-tests@ test /Users/alex/programming/javascript/todomvc/tests
> mocha allTests.js --no-timeouts --reporter spec --framework=gopherjs-humble



  TodoMVC - gopherjs-humble
    When page is initially opened
    No Todos
      ✓ should hide #main and #footer (40ms)
    New Todo
      ✓ should allow me to add todo items (397ms)
      ✓ should clear text input field when an item is added (142ms)
      ✓ should append new items to the bottom of the list (428ms)
      ✓ should trim text input (150ms)
      ✓ should show #main and #footer when items added (169ms)
    Mark all as completed
      ✓ should allow me to mark all items as completed (108ms)
      ✓ should allow me to clear the completion state of all items (145ms)
      ✓ complete all checkbox should update state when items are completed / cleared (204ms)
    Item
      ✓ should allow me to mark items as complete (372ms)
      ✓ should allow me to un-mark items as complete (370ms)
    Editing
      ✓ should hide other controls when editing
      ✓ should save edits on enter (291ms)
      ✓ should save edits on blur (308ms)
      ✓ should trim entered text (286ms)
      ✓ should remove the item if an empty text string was entered (233ms)
      ✓ should cancel edits on escape (265ms)
    Counter
      ✓ should display the current number of todo items (256ms)
    Clear completed button
      ✓ should display the correct text (79ms)
      ✓ should remove completed items when clicked (219ms)
      ✓ should be hidden when there are no items that are completed (130ms)
    Persistence
      ✓ should persist its data (1109ms)
    Routing
      ✓ should allow me to display active items (267ms)
      ✓ should respect the back button (594ms)
      ✓ should allow me to display completed items (185ms)
      ✓ should allow me to display all items (426ms)
      ✓ should highlight the currently applied filter (173ms)


  28 passing (54s)
```

I excluded repeated occurrences of the following message from the output:

```
[ { level: { name: 'SEVERE', value: 1000 },
    message: 'http://localhost:8000/favicon.ico 0:0 Failed to load resource: the server responded with a status of 404 (Not Found)',
    timestamp: 1442701209861,
    type: '',
    closure_uid_920526640: undefined } ]
```

I think it is safe to ignore, but let me know if I am supposed to include some kind of favicon.


albrow:
Travis is failing because the compiled js code is not checked into version control. Based on what the other compile-to-js examples are doing, it looks like I need to include it. I'll fix that now.

albrow:
With the compiled js included in version control, travis is timing out after receiving no output for 10 minutes. I believe what's happening is that jscs is taking a long time because the file is so big, and presumably is full of linter errors. What should I do here?

albrow:
I've added `// jscs:disable` to the top of the compiled js file, which according to [this thread](http://stackoverflow.com/questions/25223149/is-there-any-way-for-jscs-to-ignore-rules-per-file-block-or-line) will make jscs ignore the file. However, it requires jscs version >= 1.6. Opening a separate PR for this.

albrow:
Disregard the comment about jscs version. I was able to finally get the tests to pass by explicitly adding a line to .jscsrc. I think what was happening was that even with the `// jscs:disable` line at the top of the file, jscs was still parsing the entire file and looking for a `// jscs:enable`.

All the travis tests pass now. I would really love to get this included!

By the way, if you're wondering why someone might want to write frontend code in go in the first place, I wrote this [paragraph in the main README for Humble](https://github.com/go-humble/humble#why-write-client-side-code-in-go).

samccone:
Thanks for the work on this one @albrow the team will discuss. In the meantime is there someone who can review the code for best practices?

samccone:
ping @albrow ^

albrow:
@samccone sorry I didn't realize your last question was directed at me. Assuming you want someone with some go experience to review the code? I think I can find someone.

samccone:
no worries @albrow , Do you know of anyone from gopherjs that might be able to :eyes: ?

albrow:
@samccone I just asked some people in the community. I think I'll be able to find someone.

examples/gopherjs-humble/go/templates/partials/todo.tmpl
```diff
@@ -0,0 +1,8 @@
+<!-- <li {{ if .Completed }}class="completed"{{ end }}> -->
```
shurcooL: Is this HTML comment meant to be here? What is its purpose, is it to give context on how this template is used? Or is this not a comment somehow?

examples/gopherjs-humble/go/models/todo.go
```diff
@@ -0,0 +1,87 @@
...
+		Id:        todo.id,
+		Completed: todo.completed,
+		Title:     todo.title,
+	})
+}
+
+// UnmarshalJSON converts json data to a todo object.
+func (todo *Todo) UnmarshalJSON(data []byte) error {
+	jt := &jsonTodo{}
+	if err := json.Unmarshal(data, jt); err != nil {
```
shurcooL: Very minor, but this could be `var jt jsonTodo` followed by `json.Unmarshal(data, &jt)`, which _may_ be slightly better. Completely up to you, because the two are basically equivalent. I just prefer to avoid creating pointers when it's not necessary (since you can just take address of the value instead).

examples/gopherjs-humble/go/views/app.go
```diff
@@ -0,0 +1,155 @@
...
+	for _, todo := range v.Todos.Filter(v.predicate) {
+		todoView := NewTodo(todo)
+		// NOTE: the todomvc tests require that the top-level element for the
+		// todo view is an li element. Unfortunately there is no way to express
+		// this while also having template logic determine whether or not the
+		// li element should have the class "completed". I would prefer to have
+		// the top-level element for each todo be a div wrapper, and to include
+		// the li element inside the template. The workaround for now is to create
+		// the li element and set it's class manually.
+		todoView.SetElement(document.CreateElement("li"))
```
shurcooL: Oh, I think the above answers my earlier question about the commented out `<li>` in the template. Perhaps you should add a note there that points to this note to explain it.

examples/gopherjs-humble/go/views/todo.go
```diff
@@ -0,0 +1,106 @@
...
+	v.Model.Remove()
+}
+
+// Edit puts the Todo view into an editing state, changing it's appearance and
+// allowing it to be edited.
+func (v *Todo) Edit(ev dom.Event) {
+	addClass(v.Element(), "editing")
+	input, ok := v.Element().QuerySelector(".edit").(*dom.HTMLInputElement)
+	if !ok {
+		panic("Could not convert to dom.HTMLInputElement")
```
shurcooL: Is this manual handling of failed type assertions really needed here? I think you could just replace all instances of:

```Go
foo, ok := x.(y)
if !ok {
    panic("Could not convert x to y")
}
```

With the equivalent shorter version:

```Go
foo := x.(y)
```

See https://golang.org/ref/spec#Type_assertions:

> If the type assertion is false, a run-time panic occurs.

The compiler will produce pretty much the same panic message as what you've written manually. E.g.:

```
panic: interface conversion: foo is bar, not *dom.HTMLInputElement
```

Unless there's a specific reason you're choosing to go with the long form, perhaps something related to GopherJS?

shurcooL:
Hello,

I've been using Go for about 2 years and I'm collaborator for the GopherJS compiler. I've reviewed this PR and left some inline comments above.

First off, I think this is really cool! I've tried the demo at http://d3cqowlbjfdjrm.cloudfront.net/ and everything worked as one would expect from a TodoMVC app!

I've reviewed the Go code included here primarily for style issues, making sure it follows [idiomatic Go style guide](https://github.com/golang/go/wiki/CodeReviewComments), does not have `gofmt` formatting issues, `go vet` issues or `go lint` issues. In that regard, it is really well written and I did not spot any problems. It is also very well documented and very readable.

I have not used the humble packages before, but they seem well described, well organized and made sense to me. I've made note of the `localstor` package for my future needs when I want to access Local Storage API from Go programs.

Overall, I think this looks great. Most of my inline comments were minor comments, the largest being about questionable type assertion panics style. If they're addressed, this definitely has LGTM from my side.

Very nice job @albrow!

samccone:
wow thanks so much for the great review @shurcooL :clap:

albrow:
@shurcooL thanks for the review! I'll apply these suggestions as soon as I get a chance, hopefully sometime toward the end of the week.

arthurvr:
There's also a merge conflict, mind fixing that?

shurcooL:
There's also a behavior bug that we caught (in the check/uncheck all button), and I think @albrow is working on resolving that.

samccone:
:+1: thanks @shurcooL and @albrow 

I am personally super excited for this to land :tada:

albrow:
My other computer got hit with some BSODs this past weekend, so that took up most of my time :/ Planning on applying these changes tonight!

examples/gopherjs-humble/go/views/todo.go
albrow: No, you're right. I just did this out of habit since you usually want to return an error instead of panicking :) In this case there's no caller to return to so panicking makes sense. I'll change it.

albrow:
Okay I've addressed all of @shurcooL's feedback (thanks again!) and fixed the merge conflict. Should be good to go I think.

By the way, I wanted to elaborate on the issue @shurcooL mentioned with the toggle-all button. My original implementation was broken, but the tests didn't catch it. Here's how it behaved before:

Broken toggle-all

And here is the corrected version:

Fixed toggle-all

This is probably something we should consider adding to the tests, as I can imagine it is a fairly common mistake.

shurcooL:
> This is probably something we should consider adding to the tests, as I can imagine it is a fairly common mistake.

I fully agree. I was extremely surprised such basic app functionality (all it takes is starting state + 1 action to get incorrect result state) was not caught by one of the many tests.

samccone:
Hi @albrow I ran your implementation through my automated memory leak detection tool and it looks like you have a memory leak in your application.

screen shot 2015-10-05 at 10 35 33 pm

Over 5 runs you have
* +30 node allocations
* `<a heap size increase`>
* +5 JS event listeners


you can reproduce this locally via

> git clone git@github.com:samccone/drool.git
> git checkout sjs/taste

Apply this patch

```patch
diff --git a/taste.js b/taste.js
index cc406c0..ba07de7 100644
--- a/taste.js
+++ b/taste.js
@@ -33,11 +33,11 @@ function listSelector(name) {
   });
 }
 
-["examples/angularjs_require",  "examples/flight", "examples/react", "examples/canjs", "examples/backbone", "examples/angularjs", "examples/emberjs", "examples/knockoutjs", "examples/dojo", "examples/mithril", "examples/ampersand", "examples/vue", "examples/backbone_marionette", "examples/troopjs_require", "examples/spine", "examples/vanilladart/build/web", "examples/closure", "examples/elm", "examples/angular-dart/web", "examples/typescript-backbone", "examples/typescript-angular", "examples/serenadejs", "examples/reagent", "examples/thorax", "examples/chaplin-brunch/public", "examples/backbone_require", "examples/knockoutjs_require","examples/canjs_require", "examples/thorax_lumbar/public", "examples/somajs_require", "examples/durandal", "examples/lavaca_require", "examples/cujo/index.html", "examples/sammyjs", "examples/epitome", "examples/somajs", "examples/duel/www", "examples/kendo", "examples/puremvc", "examples/olives", "examples/dijon", "examples/rappidjs", "examples/extjs_deftjs", "examples/enyo_backbone", "examples/angularjs-perf", "examples/sapui5", "examples/exoskeleton", "examples/atmajs", "examples/ractive", "examples/react-backbone", "examples/aurelia", "examples/foam", "examples/webrx", "examples/angular2"].reverse().forEach(function(name) {
+["examples/gopherjs-humble"].reverse().forEach(function(name) {
   drool.flow({
     repeatCount: 5,
     setup: function() {
-      driver.get('http://todomvc.com/' + name);
+      driver.get('http://localhost:7777/' + name);
     },
     action: function(name) {
       driver.wait(function() {
```


* Start your webserver on port 7777
* run node taste.js

----------

I manually tested this and profiled using chrome debugger and found the following:

##### Steps to reproduce

* start timeline profiler
  * add three todos
  * delete three todos
* (repeat above 2 more times)
* force GC
* stop timeline profiler

Here is the timeline after forcing a GC (you can see things do not drop back down to 0)

screen shot 2015-10-05 at 10 40 55 pm

Here is the profiler with heap snapshots .. taking diffs between my actions

screen shot 2015-10-05 at 10 44 59 pm

You can see that there is something in `g_fillreqdocs` that seems to be retaining :fountain:

arthurvr:
You forgot to add the app to learn.json and the index.html file (the todomvc.com site), can you take care of that? :)

Thanks for the PR! This is really neat work!

albrow:
@samccone wow thanks for putting together that awesome report. I'll take a look at the memory leaks as soon as I can.

@arthurvr yes, I'll do that.

I also noticed there is another merge conflict so I'll fix that too.

samccone:
Let me know if you have any questions @albrow :)

albrow:
Okay I added to learn.js and index.html, fixed the merge conflict, and was able to fix part of the memory leak. There are no more left over event listeners, but I'm still trying to track down the dangling DOM node references.

@samccone current output of running your automated memory leak detector is:

```
examples/humble, 30, 283912, 0
```

The heap size difference is never the same between runs and sometimes is in the negatives, like:

```
examples/humble, 30, -3738512, 0
```

Can I safely assume that it's the other two numbers that really matter?

albrow:
I still can't figure out what is causing the DOM node reference leak. I have tried changing a ton of different things to see what might be causing it, including:

1. Adding/removing event listeners to views/app.go and views/todo.go
2. Adding/removing various DOM elements to the templates
3. Adding/removing routes
4. Creating a completely new app view and restarting the router on each render

None of the above had any effect (either an increase or decrease) in the number of node references.

I'm not sure what `g_fillreqdocs` is. It does not exist in the code I wrote. I also could not find it in the source code for GopherJS or the DOM bindings.

@samccone do you have any tips or suggestions for tracking this down?

samccone:
hey @albrow Ill check this out and run through it again, I fear that depending on how your inputs are being rendered, you might be hitting https://code.google.com/p/chromium/issues/detail?id=516153

But I will check


As far as the heap size, yeah do not worry about it, it is a highly variable number depending on a bunch of things :)

samccone:
Hi @albrow I did a bunch more digging into this leak.

Ok a simple way for you to reproduce locally is just to run the profiler and add / remove 20+ todos and then click the garbage can icon a few times and then stop the profiler.

That results in this:

screen shot 2015-10-12 at 8 50 32 pm

A pretty significant increate in the node count




From there I did some digging around the heap snapshot to see if I could figure out what was retaining  DOM nodes in the tree

There was some really curious stuff going on here:

screen shot 2015-10-12 at 8 30 40 pm

Which terminated to the root of the doc here:

screen shot 2015-10-12 at 8 30 06 pm

I kept going and I think*** what is leaking is showing up under the Array / Typ constructor
screen shot 2015-10-12 at 9 03 13 pm

screen shot 2015-10-12 at 9 02 26 pm

screen shot 2015-10-12 at 9 06 40 pm

With a "parent" of **methodSetCache**

screen shot 2015-10-12 at 9 04 23 pm


--------

Well hopefully this is helpful to you, or at least makes it clear how to reproduce the node leak

Thanks :sparkles:

shurcooL:
The `$externalizeWrapper` line points to [dom.go:1312](https://github.com/dominikh/go-js-dom/blob/a900f4636e01f006b98f184aaba35f669bcf568f/dom.go#L1312).

@albrow, one of the things worth trying IMO when investigating this is skipping the `AddEventListener` wrapper of `dom` package and using the raw js calls, i.e. `js.Global.Get("myButton").Call("addEventListener", "click", func() { ...`. Just to see if skipping that wrapper has any effect here. It should be easy to try.

I'll /cc @dominikh and @neelance since they're familiar with the relevant code and might be interested in learning the outcome here.

samccone:
awesome :)  thanks @shurcooL, let me know if I can be of __any__ more help here :)

samccone:
@shurcooL Any update on this :)

albrow:
@shurcooL I should have time to try out your suggestion of skipping the `AddEventListener` wrapper in the `dome` package this weekend. My hunch is that this will not have any effect. I already tried adding/removing event listeners and did not observe any increase or decrease in the number of leaked node references. However, it is possible that the references are only leaked for a certain kind of listener or a listener on a certain kind of element.

samccone:
any news @albrow :)

albrow:
I'm sorry for the late reply. I was on vacation and then came down with a cold. For some reason, without changing the code at all, when I ran the automated leak detection tool today I got:

```
examples/humble, 26, 3829096, 0
```

So there are now 26 leaking dom node references instead of 30, which we were seeing before. I did update the GopherJS compiler which may have had something to do with it. The other thing that may have changed is the version of Chrome since it autoupdates.

I'll update you with any new information I find today.

albrow:
@samccone one of the objects you found in your report, the one with properties `ForceHashURL` and `ShouldInterceptLinks`, is the router object from [`github.com/go-humble/router`](https://github.com/go-humble/router). Since it was one of the few things I recognized as code I wrote, I thought it may have been the cause. However, I tried removing all router-related code and it made no difference in the number of leaked nodes.

I have also begun replacing all event listeners with raw js calls as @shurcooL suggested. It's tedious work and I only had time to replace a few listeners. So far it has made no difference. It's still possible that the leak is dependent on the type of listener or the type of element, so I'll replace all the event listeners before drawing a conclusion.

I have to go for now but may be able to do some more investigating tonight.

samccone:
Hey @albrow thanks for digging in... so I did some more research tonight, and decided to look at this another way.

What if we started with where the event listeners are added and then worked our way from there.. so I recorded a timeline and added and removed a few nodes. I then zoomed in super far to see the methods that we allocating new listeners and DOM nodes --- basically it all comes down to this method :point_down: 

screen shot 2015-11-02 at 10 29 25 pm


What if we starting pulling apart this method and also the corresponding remove method :point_down: 

screen shot 2015-11-02 at 10 37 01 pm


Here are a few questions that I would ask....

What if you did not set any event listeners other than the delete button? What does that do to the code and the DOM node count? 

What if you do not unbind on purpose? how bad do things get? 

How stripped down can you make this test case?


> On a side note I tried to run edit some things and compile however I got

`exec: "temple": executable file not found` when running `go generate ./...`

:cry:

shurcooL:
> On a side note I tried to run edit some things and compile however I got
>
> `exec: "temple": executable file not found` when running `go generate ./...`

I can help with this. It sounds like `temple` binary is not in your GOPATH, but it's needed for that go generate directive.

You can download and install `temple` from https://github.com/go-humble/temple#installation:

```
go get -u github.com/go-humble/temple
```

That installs the `temple` binary into your `$GOPATH/bin`. If that's in your `$PATH`, you should be good to go. You can learn more about `go generate` command [here](https://golang.org/cmd/go/#hdr-Generate_Go_files_by_processing_source), but keep in mind it's meant to be used by developers of a package, not its users.

albrow:
@samccone that's my fault for not listing temple as a dependency in the README. I'll add it as soon as I get a chance.

@shurcooL thanks for stepping in to help :)

albrow:
@samccone I had some time to do some more digging and I can answer your questions:

> What if you did not set any event listeners other than the delete button? What does that do to the code and the DOM node count?

I tried this and there was no effect on the number of leaked nodes.

> What if you do not unbind on purpose? how bad do things get?

I used the raw js bindings to add the event listener and did not explicitly unbind. @shurcooL you were also interested in this result? Again, there was no effect.

(The reason I did the unbinding in the first place was to fix the leaked event listeners. I didn't expect this to be required and I'm still questioning whether it is. It did fix the leaked event listeners, but those may have been occurring *because* of the leaked node references.)

> How stripped down can you make this test case?

That is tbd. I'm going to focus now on creating a minimal test case to see if this is reproducible.

I also added instructions to the README for installing temple :)

albrow:
I am now fairly convinced that this is due to a bug in chrome. I was able to reproduce the node reference leak with a smaller go application. Then I rewrote that application in pure javascript and observed the same memory leak. Eureka!

The leak only occurs on input elements. To reproduce the leak, I rendered an input element, attached an event listener to it, and then replaced the input element. Here is the javascript code:

``` javascript
var body = document.querySelector("body");
render();

function render() {
	body.innerHTML = "<input />";
	var input = body.querySelector("input");
	input.focus();
	input.addEventListener("keypress", function(ev) {
		// 13 is the code for enter
		if (ev.keyCode === 13) {
			render();
		}
	})
}
```

The basic idea behind the code is that each time you press enter, it renders a completely new input element and attaches a new event listener to it. Because there are no more references to the old input element, the element itself and the event listener *should* be garbage collected.

Here is a demonstration of the leak using chrome developer tools:

Node reference leak in pure javascript

The gif is a screen recording of me doing the following:

1. Starting the timeline profiler.
2. Forcing garbage collection a few times.
3. Typing in the input and pressing enter a few times to cause it to re-render.
4. Forcing garbage collection again a few times.
5. Stopping the timeline profiler.

As you can see, at the start there were 87 nodes and at the end there were 98.

According to this [stackoverflow answer](http://stackoverflow.com/questions/12528049/if-a-dom-element-is-removed-are-its-listeners-also-removed-from-memory), event listeners are supposed to be automatically cleaned up if the element they are attached to is deleted. I have observed that is indeed what happens if you use, say, an anchor element instead of an input element in the above code. In that case, there is no memory leak. As far as I can tell, it only happens with input elements.

For reference, here is the go code that produces the same memory leak:

``` go
package main

import (
	"fmt"

	"honnef.co/go/js/dom"
)

var (
	document = dom.GetWindow().Document()
	body     = document.QuerySelector("body")
)

func main() {
	render()
}

func render() {
	body.SetInnerHTML("<input />")
	input := body.QuerySelector("input").(*dom.HTMLInputElement)
	input.Focus()
	input.AddEventListener("keypress", true, func(ev dom.Event) {
		keyEvent, ok := ev.(*dom.KeyboardEvent)
		// NOTE: 13 == enter
		if ok && keyEvent.KeyCode == 13 {
			render()
		}
	})
}
```

One question which remains is why doesn't this happen with other implementations of TodoMVC? I'm not entirely sure, but it might have to do with the way my implementation renders. Each time the model changes (e.g. when you add/remove a todo), the bulk of the application is re-rendered by setting the `innerHTML`. (Obviously not the most efficient thing, but the performance is fine for this use case, even on mobile.)

@samccone what do you think of this analysis? Is there anything else you would like to see?

shurcooL:
Nice find.

Is it possible to try the test on different browsers, like Safari or Firefox? Also worth giving a shot to another version of Chrome. For example, try Chrome Canary (the latest bleeding edge development version; it can run side-by-side with normal Chrome).

samccone:
Ah the input leak ... Yeah this is a known bug due to how the undo buffer in chrome works. I was not sure how your input was being rerendered but this all makes sense ... I will give this a once over again but hopefully we can land this! Thanks a million for the work

albrow:
Just tried it out in Chrome Canary and saw the same issue. I thought I had ruled out the input leak when I added an extra input element and observed no change. But I realize looking back that it doesn't happen with just any input element; it needs to be added and then removed from the dom in order for the leak to be apparent. I'm fixing the merge conflicts now. I would also like an opportunity to look over one more time before we merge this.

albrow:
@samccone If I'm interpreting the TravicCI results correctly, the tests are failing because of the node count leak. 

``` bash
humble, 26, 260320, 0
/home/travis/build/tastejs/todomvc/tests/node_modules/selenium-webdriver/lib/goog/async/nexttick.js:41
  goog.global.setTimeout(function() { throw exception; }, 0);
                                      ^
Error: Node Count leak detected!
```

Maybe we can we make the leak detector less strict?

samccone:
yep in this case @albrow I will add a framework specific exception :) ... will do this tonight

samccone:
:boom: #1513 is working... landing this now

thanks thanks thanks @albrow 

+ :100: :100: :100: :100: :100: :100: :100:  :rocket: :guitar:  

photo on 11-9-15 at 10 10 pm

shurcooL:
Congrats! :confetti_ball: :star2:

albrow:
Awesome! Thanks for all your help @samccone and @shurcooL. Excited to be a part of the official project :smile:

PR #1442, opened 2015-09-20
Bump jscs version to 1.6

albrow:
See https://github.com/tastejs/todomvc/pull/1441#issuecomment-141721812. jscs 1.6 supports ignoring certain js files via a special comment: `// jscs:disable`. This is important for Humble and other compile-to-js frameworks.

### DIFF
--- a/tooling/package.json
+++ b/tooling/package.json
@@ -4,6 +4,6 @@
     "lint": "jscs"
   },
   "dependencies": {
-    "jscs": "^1.13.1"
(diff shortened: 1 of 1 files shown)


arthurvr:
Why do you want to downgrade the version? If a feature is in 1.6, why isn't it in 1.13?

albrow:
Oh damn. My mistake. Completely misread the version as 1.1.3. Closing this.

PR #1448, opened 2015-09-26
fix checkbox bug in the mithril TodoMVC example

nordfjord:
The bug is caused by lack of keying.

To reproduce the bug:
1. Open the app
2. Create at least two tasks
3. Complete all tasks
4. Go to the completed view
5. uncheck the top item
6. Voilá, see how the first item in the list now has an unchecked checkbox

### DIFF
--- a/examples/mithril/js/models/todo.js
+++ b/examples/mithril/js/models/todo.js
@@ -2,9 +2,17 @@
 /*global m */
 var app = app || {};
 
+var uniqueId = (function () {
+	var count = 0;
+	return function () {
+		return ++count;
+	};
+}());
+
 // Todo Model
 app.Todo = function (data) {
 	this.title = m.prop(data.title);
 	this.completed = m.prop(data.completed || false);
 	this.editing = m.prop(data.editing || false);
+	this.key = uniqueId();
 };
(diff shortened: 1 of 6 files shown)


samccone:
Thanks a ton of the PR, looks likes few lint errors on your branch. Mind taking a look?

nordfjord:
Sure thing, I think it should be fixed now, (I squashed the commits per the pull request guidelines).

However there seems to be a test failing, which I'm going to attribute to a badly written test.
I'll get on fixing the test.
The test error says the element should be hidden, and complains that it can't be found (ringing any irony bells?)

samccone:
Hi @nordfjord you are correct it looks like a bad assumption on the tests part

The best test is located here
https://github.com/tastejs/todomvc/blob/aeb5e7e2ce8a6a918ce6b9020156f126baae9d76/tests/testOperations.js#L30-L34

Let me know if run into any troubles fixing the test, I am more than willing to fix it, however I very much appreciate you volunteering to fix it! Thanks a ton!!!

nordfjord:
I seem to have figured this out,

I added an error callback to the check, which then asserts the error code is 7 (7 is the error code for element not found).
and everything seemed to run dandily on my end.

I however would like to skip squashing this commit since it's completely unrelated to the other one, so I would rather like to be able to see in the commit history that both a change to mithril was made and to the tests.

Regards,
Einar

samccone:
awesome this looks great @nordfjord awesome work :space_invader:

PR #1451, opened 2015-09-29
Add a new library: Backscatter (a Backbone extension)

tweinfeld:
Backscatter is a small Backbone extension that report events from every model/collection below a root model/collection. Backscatter reports only relevant events from entities which are active members of the model tree and ignores events from entities which "fell off" that tree, so users can freely add and remove entities without worrying about hooking them up, propagating events, clearing event handlers and referencing to discarded entities.

Backscatter can be harnessed to trigger `React.render` cycles, which makes it a great `React` companion.

This PR includes a TodoMVC demo that consists mostly of an original Backbone + React implementation. Backscatter is used in `controller/application` (one occurrence throughout the application) to deliver global change notifications and drives React.


arthurvr:
Thanks for your suggestion :balloon: 

We discussed this and decided to wait till its user base grows a bit. When you feel like it needs reconsideration please let us know. Thanks again :)

tweinfeld:
Thank you for considering Backscatter. I hope to meet here again soon :)

PR #1453, opened 2015-10-01
polymer fixups

samccone:
R: @arthurvr


arthurvr:
Can you remove the blank lines at the end of these files? Besides from that, LGTM. Thanks :dancer:

addyosmani:
Let's also loop in @ebidel on these reviews to have another pair of eyes on our Polymer implementation updates. 

In this case, looks fairly straight-forward however I'm unsure why we're breaking out the scripts to separate files. @samccone?

ebidel:
Can we hold of on this? I have a large or coming that's a bit more polymeric. It's based on the 1.0 port started by Sam. As a canonical example I think we should go with that.

arthurvr:
:+1:

addyosmani:
@ebidel SGTM. @samccone let's close this once the other PR comes in and coordinate on making the other one canonical as far as the Polymer updates go.

samccone:
> In this case, looks fairly straight-forward however I'm unsure why we're breaking out the scripts to separate files given the majority of official elements are inlined

To ensure our linting runs over all the files


> Can we hold of on this? I have a large or coming that's a bit more polymeric

Ok we can :skull: this

PR #1454, opened 2015-10-01
Example: Meteor - Iron Router

redice44:
## Why Iron Router?
Iron:Router provides routing control structures to Meteor. This example demonstrates well organized file structure for Meteor, simple routing, controllers, and controller inheritance as its enhancement over the existing Meteor example.

Try the [Live Demo](http://iron-router-todomvc.meteor.com/) hosted on [Meteor.com](https://www.meteor.com/).


samccone:
@redice44 If this is the community established best practice then we should PR this into the main meteor example. 

Currently we are trying to keep these sub permutations under control. Feel free to add this example directly to the wiki though.

Thanks a bunch for your time and the PR.

PR #1368, opened 2015-07-08
👾 Add proof of concept leak test 👾 

samccone:
I know this has been a longish time coming but here it is.
Ideally this will run in the same way that we run integration tests, based on the git diff.

This perf test is inline with all of my perf testing in that I have essentially codified the process here. 

https://github.com/samccone/leaky provides us with the instrumentation layer so we do not have to worry about it and get to reuse all of our selenium work in our integration suite :dancer: .

I currently have this hard coded to test against react, just to get some feedback on what everyone is  thinking.

------------

:fountain:  **React**:

```
initial snapshot: {"documents":3,"jsEventListeners":12,"jsHeapSizeUsed":9127312,"nodes":202}
end snapshot:     {"documents":3,"jsEventListeners":12,"jsHeapSizeUsed":10889192,"nodes":332}
```

node leak ^ as expected via https://github.com/tastejs/todomvc/issues/1332#issuecomment-115057932

:fountain: **Vue**

```
initial snapshot: {"documents":4,"jsEventListeners":7,"jsHeapSizeUsed":4696168,"nodes":293}
end snapshot:     {"documents":4,"jsEventListeners":7,"jsHeapSizeUsed":5681776,"nodes":383}
```

node leak ^ as expected via https://github.com/tastejs/todomvc/issues/1354

:fountain: **Angularjs**

```
initial snapshot: {"documents":3,"jsEventListeners":8,"jsHeapSizeUsed":6224976,"nodes":594}
end snapshot:     {"documents":3,"jsEventListeners":8,"jsHeapSizeUsed":8055816,"nodes":684}
```

node leak ^ as expected via https://github.com/tastejs/todomvc/issues/1335

:fountain: **Knockoutjs**

```
initial snapshot: {"documents":3,"jsEventListeners":7,"jsHeapSizeUsed":4346992,"nodes":273}
end snapshot:     {"documents":3,"jsEventListeners":37,"jsHeapSizeUsed":5414728,"nodes":363}
```

node and listener leak ^ as expected via https://github.com/tastejs/todomvc/issues/1334

:fountain: **Ember**

```
initial snapshot: {"documents":3,"jsEventListeners":5,"jsHeapSizeUsed":15816152,"nodes":267}
end snapshot:     {"documents":3,"jsEventListeners":5,"jsHeapSizeUsed":17672328,"nodes":517}
```

node leak ^ as expected via https://github.com/tastejs/todomvc/issues/1333

----------

*^ is by no means a comprehensive run of all of the implementations just the ones I have run*

PR #1464, opened 2015-10-09
⚡️ Automated Leak Detection on CI ⚡️ 

samccone:
With this PR [Drool](https://github.com/samccone/drool)  is now providing us with the ability to fail CI when an implementation has memory leaks.

The perf tests are now run on every PR :space_invader: :spaghetti: 

1. Add automated leak detection :fountain: 
2. Only run deploy code paths on deploy :) (faster test runs for everyone) :horse: 
3. Add some more compiled src ignores

### DIFF
--- a/tests/framework-path-lookup.js
+++ b/tests/framework-path-lookup.js
@@ -1,0 +1,62 @@
+var fs = require('fs');
+var frameworkNamePattern = /^[a-z-_\d]+$/;
+
+var excludedFrameworks = [
+	// this implementation deviates from the specification to such an extent that they are
+	// not worth testing via a generic mechanism
+	'gwt',
+	// these implementations cannot be run offline, because they are hosted
+	'firebase-angular', 'meteor', 'socketstream',
+	// YUI is a special case here, it is not hosted, but fetches JS files dynamically
+	'yui',
+	// these frameworks take a long time to start-up, and there is no easy way to determine when they are ready
+	'cujo',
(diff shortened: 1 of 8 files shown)


samccone:
:green_apple:  CI is working (angular2 fails because of no routing)

https://travis-ci.org/tastejs/todomvc/builds/84436464

samccone:
xref #1368

ColinEberhardt:
Wow, that is awesome! I'll definitely recommend using Drool to the developers where I work.

passy:
I love this! Just tested this with a few apps and it's sooo good! Thanks for this, @samccone!

tests/framework-path-lookup.js
```diff
@@ -0,0 +1,62 @@
...
+	list = list.filter(function (framework) {
+		return excludedFrameworks.indexOf(framework.name) === -1;
+	});
+
+	return list.filter(function (framework) {
+		return [].concat(names).some(function (f) {
+			return f === framework.name;
+		});
+	});
+}
```
arthurvr: Missing semicolon, why didn't the linting process catch this?

passy: Or me! Outrageous! Fixed in master with 1567e60.

arthurvr: Thanks @passy! Still we should check what's wrong with the linting tooling though.

PR #1467, opened 2015-10-11
use express.js server for tests; closes #1465

### LINKED ISSUE #1465: Lack of favicon.ico in root directory raises 404 error, cluttering test output
When I run the tests per the instructions in `tests/README.md`, I get a 404 error for favicon.ico in many/most/all(?) of the tests.

Example:

```
  TodoMVC - backbone
    When page is initially opened
[ { level: { name: 'SEVERE', value: 1000 },
    message: 'http://localhost:8000/favicon.ico 0:0 Failed to load resource: the server responded with a status of 404 (Not Found)',
    timestamp: 1444529718462,
    type: '',
    closure_uid_682459046: undefined } ]
    No Todos
      ✓ should hide #main and #footer (131ms)
[ { level: { name: 'SEVERE', value: 1000 },
    message: 'http://localhost:8000/favicon.ico 0:0 Failed to load resource: the server responded with a status of 404 (Not Found)',
    timestamp: 1444529720273,
    type: '',
    closure_uid_682459046: undefined } ]
...
```

The tests pass, but these error messages terribly clutter up the test output.

I imagine that there is some setting to suppress the printing of the errors, but that is probably not ideal because it would be nice to see other, legitimate errors in the test output.

The problem can be fixed by simply creating in the root directory a symlink to the favicon.ico that is in the `site-assets` directory:

```sh
~/Downloads/todomvc $ ln -s site-assets/favicon.ico favicon.ico
```

Now we get a much neater test output:

```
  TodoMVC - backbone
    When page is initially opened
    No Todos
      ✓ should hide #main and #footer (98ms)
...
```

I'd make a pull request adding the symlink, but I feel like I am probably missing something? Otherwise I imagine that someone else would have raised this issue by now. Apologies if that is the case. Additionally, I'm not sure if y'all want that symlink cluttering up the root directory.

davidrunger:
Use an express.js server for tests instead of http-server. The primary
reason for this is that express has an available middleware that allows
us to specificy the location of the favicon other than the root
directory. This is needed because the favicon lives in the site-assets
directory.

### DIFF
--- a/tests/package.json
+++ b/tests/package.json
@@ -4,7 +4,6 @@
   "private": true,
   "devDependencies": {
     "chromedriver": "^2.18.0",
-    "http-server": "^0.7.5",
     "mocha": "*",
     "mocha-known-issues-reporter": "git+https://github.com/ColinEberhardt/mocha-known-issues-reporter.git#v0.0.0",
     "optimist": "^0.6.1",
@@ -12,7 +11,6 @@
     "drool": "0.2.2"
   },
   "scripts": {
-    "serve": "http-server -p 8000 ..",
     "test": "mocha allTests.js --no-timeouts --reporter spec"
   }
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,10 +30,6 @@ before_install:
     - sh -e /etc/init.d/xvfb start
     - curl -Lo chrome.zip https://download-chromium.appspot.com/dl/Linux_x64 && unzip chrome.zip
     - curl -Lo chromedriver.zip http://chromedriver.storage.googleapis.com/2.18/chromedriver_linux64.zip && unzip chromedriver.zip
-before_script:
-    - cp site-assets/favicon.ico favicon.ico
-    - python -m SimpleHTTPServer > /dev/null 2>&1 &
-    - sleep 2
 script:
     # We want to gate on passing tests and a successful build
     - ./test-runner.sh
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -109,3 +109,7 @@ gulp.task('pagespeed', pagespeed.bind(null, {
 gulp.task('serve', function (cb) {
 	app.listen(8080, cb);
 });
+
+gulp.task('test-server', function (cb) {
+	app.listen(8000, cb);
+});
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "gulp-vulcanize": "^1.0.0",
     "jshint-stylish": "^1.0.0",
     "psi": "^0.1.1",
-    "run-sequence": "^1.0.1"
+    "run-sequence": "^1.0.1",
+    "serve-favicon": "^2.3.0"
   }
 }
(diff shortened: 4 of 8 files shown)


tests/package.json
```diff
@@ -12,7 +11,6 @@
     "drool": "0.2.2"
   },
   "scripts": {
-    "serve": "http-server -p 8000 ..",
```
passy: What do you think about making this an alias for the `gulp test-server`?

passy:
Looks good to me! :+1: 

Waiting for @samccone to give his okay. :)

tests/package.json
davidrunger: Well, the sort of nice thing about `gulp test-server` is that one doesn't have to be in the `tests` directory when executing that command. So we can say "Start the test server using `gulp test-server`" instead of "Start the test server by navigating to `tests` and running `npm run serve`". Plus, if we have `npm run serve` as an alias like that, it adds a small extra step in the chain that someone has to follow if they want to see what's happening with that command. For these minor reasons, I favor having `gulp test-server` at least be the primary command for running the test server (i.e. the command that is provided in README.md). However, if you are thinking more along the lines of leaving `npm run serve` as an undocumented legacy command in order to not break the API for people who have already become accustomed to using that command, then I'm down with that and would be happy to squash that change into this pull request.

sindresorhus:
:+1:

samccone:
Hey @davidrunger thanks for the PR

a few more things that I would like to see changed before we merge this in.

1. we no longer need to do this work for CI https://github.com/tastejs/todomvc/blob/master/.travis.yml#L34-L36
2. Move starting up the local server into here https://github.com/tastejs/todomvc/blob/master/tests/run.sh or into https://github.com/tastejs/todomvc/blob/master/test-runner.sh#L35
3. Make a tmp change to a framework example just so we can test everything works on CI

thanks a ton @davidrunger

davidrunger:
Good catches, @samccone.

Hopefully this pull request is now the way you'd like it. As per your point # 3, I made a pull request with a breaking change to a framework (Backbone), and the Travis CI tests ran and caught the integration test failures, as hoped/expected. That pull request is here: https://github.com/tastejs/todomvc/pull/1470

This pull request (thanks to `git push -f`) is now exactly the same as #1470, minus the breaking change to Backbone. So it should be good to merge, as far as I can tell.

davidrunger:
Alright, @samccone, third time's the charm, hopefully. :) I humbly submit my latest attempt, now (re)featuring `sleep 2`.

FWIW, I performed [another test build](https://travis-ci.org/tastejs/todomvc/builds/84993870), breaking the Backbone example, and everything still seems to be working on Travis.

tests/package.json
samccone: :+1: your reasoning is sound to me @davidrunger :)

samccone:
in we go :) thanks @passy for the :eyes: 

Thanks sooo much @davidrunger for the work :hammer: :wrench:

davidrunger:
Happy to help. :) Thank you @samccone for your time, your attention to detail, and your patience.

PR #1470, opened 2015-10-11
Travis CI test. Don't merge. Testing using express.js server for tests.

davidrunger:
This is a temporary version of this pull request so that we can test that everything is working correctly with Travis CI now using a different server to run the tests. The Travis CI build should produce a couple of integration test errors because of the breaking change to `examples/backbone/js/views/app-view.js`.


davidrunger:
Works as expected. :+1:

PR #1473, opened 2015-10-13
new app typescript-react

remojansen:
Sorry, I messed up the previous pull request (https://github.com/tastejs/todomvc/pull/1469) when squashing...

## Original pull-request description
Hi,

I have created a typescript-react app. There are two typescript examples (backbone and angular) but working with react is quite different. Since typescript 1.6 a new type of file is available (`.tsx` instead of `.ts`). The TSX files are a typed superset of JSX files and the example showcase how to work with this kind of files. 

The example also showcases how to work with a classical object oriented approach (`SomeClass extends React.Component<IProps, IState>` instead of `React.createClass({/* ... */})`), which is more common in TypeScript apps.

Please find below the output of the unit tests.

### Tests output
```

> todomvc-browser-tests@ test /Users/remojansen/Desktop/todomvc/tests
> mocha allTests.js --no-timeouts --reporter spec --framework=typescript-react



  TodoMVC - typescript-react
    When page is initially opened

      ✓ should focus on the todo input field
[ { level: { name: 'SEVERE', value: 1000 },
    message: 'http://localhost:8000/favicon.ico 0:0 Failed to load resource: the server responded with a status of 404 (Not Found)',
    timestamp: 1444579805995,
    type: '',
    closure_uid_747125925: undefined } ]
    No Todos

      ✓ should hide #main and #footer (70ms)
[ { level: { name: 'SEVERE', value: 1000 },
    message: 'http://localhost:8000/favicon.ico 0:0 Failed to load resource: the server responded with a status of 404 (Not Found)',
    timestamp: 1444579807716,
    type: '',
    closure_uid_747125925: undefined } ]
    New Todo

      ✓ should allow me to add todo items (607ms)
[ { level: { name: 'SEVERE', value: 1000 },
    message: 'http://localhost:8000/favicon.ico 0:0 Failed to load resource: the server responded with a status of 404 (Not Found)',
    timestamp: 1444579809729,
    type: '',
    closure_uid_747125925: undefined },
  { level: { name: 'WARNING', value: 900 },
    message: 'http://localhost:8000/examples/typescript-react/node_modules/react/dist/react-with-addons.js 21628:15 Warning: React.addons.classSet will be deprecated in a future version. See http://fb.me/react-addons-classset',
    timestamp: 1444579809951,
    type: '',
    closure_uid_747125925: undefined } ]

      ✓ should clear text input field when an item is added (276ms)
[ { level: { name: 'SEVERE', value: 1000 },
[...]

PR #1474, opened 2015-10-14
Fix for react addons clasSset deprecated warning

remojansen:
As [requested](https://github.com/tastejs/todomvc/pull/1473#issuecomment-147844425) by @samccone this pull request fixes the `Warning: React.addons.classSet` in React example.


passy:
Thank you! :star:

PR #1480, opened 2015-10-18
scalajs-react example: Update to new API

elacin:
Hey again people :)

There has just been a release of scalajs-react with a big [update](https://github.com/japgolly/scalajs-react/blob/master/doc/CHANGELOG-0.10.md#unified-the-two-ways-of-writing-callbacks) of the API for doing callbacks, so I updated the todomvc example to match it.

As a consequence of the unified callback api, we no longer need the scalaz dependency, so that is taken out. React itself is updated from 0.12 to the recently released 0.14


examples/scalajs-react/src/main/scala/todomvc/CTodoList.scala
```diff
@@ -31,34 +33,45 @@ object CTodoList {
...
-    def updateTitle(id: TodoId)(title: Title): IO[Unit] =
-      editingDone(cb = $.props.model.update(id, title))
+    /**
+     *  A backend lives for the entire life of a component. During that time,
+     *   it might receive new Props,
+     *   so we use this mechanism to keep state that is derived from Props, so
+     *   we only update it again if Props changed in a meaningful way (as determined
+     *   by the implicit `Reusability` defined above )
+     */
+    case class Callbacks(P: Props){
```
passy: Nit: Space before opening curly :)

examples/scalajs-react/src/main/scala/todomvc/CTodoList.scala
```diff
@@ -31,34 +33,45 @@ object CTodoList {
...
+      val toggleAll: ReactEventI => Callback =
+        e => P.model.toggleAll(e.target.checked)
+    }
 
-    def startEditing(id: TodoId): IO[Unit] =
-      $.modStateIO(_.copy(editing = Some(id)))
+    val cbs = Px.cbA($.props).map(Callbacks)
+
+    val startEditing: TodoId ⇒ Callback =
+      id ⇒ $.modState(_.copy(editing = Some(id)))
```
passy: Mixing unicode and ASCII arrows here. Could you only use `=>` here for easier editability (is that even a word?)?

examples/scalajs-react/src/main/scala/todomvc/CTodoList.scala
```diff
@@ -70,62 +83,64 @@ object CTodoList {
           <.input(
             ^.className     := "new-todo",
             ^.placeholder   := "What needs to be done?",
-            ^.onKeyDown  ~~>? handleNewTodoKeyDown _,
+            ^.onKeyDown   ==>? cbs.value().handleNewTodoKeyDown,
```
passy: Just wondering, is it necessary/correct/good style to call `value()` as opposed to treating it as a value -> `value`?

passy:
Thanks for the update! Good to see an update to React 0.14 so quickly. Left a few comments, but CI is already green so I'm happy to merge this as soon as those are addressed. :)

examples/scalajs-react/src/main/scala/todomvc/CTodoList.scala
elacin: i fully approve of that as a word. Yeah, this is the moment when i tell my IDE to stop autounicodifying those arrows (how about that for a word?)

passy: Haha, I know that very well. :)

examples/scalajs-react/src/main/scala/todomvc/CTodoList.scala
elacin: Yeah, i admit it looks a bit bad, but it's nothing like calling `Option.get` for example - it's completely safe. What it is meant to signal to the reader is that calling `value()` might trigger a new `Callbacks` instance to be created if, in this case, `Props` has changed in a meaningful way since last time it was accessed. I think i will add a comment in the source to make it clearer.

passy: I see! I didn't realize that this had more effects than just retrieving the value from the state. In that case, this is totally justified.

elacin:
Yeah my pleasure - it's great that you people are responsive with PRs here. 
Fixed the things you commented on, and added that comment :)

arthurvr:
LGTM. Thanks for the updates!

PR #1481, opened 2015-10-19
React: Remove direct DOM manipulation

passy:
Turn the `<input>` into a
[Controlled Component](https://facebook.github.io/react/docs/forms.html#controlled-components)
and avoid fiddling with the DOM in an imperative way.

[Review on Reviewable](https://reviewable.io/reviews/tastejs/todomvc/1481)

### DIFF
--- a/examples/react/js/app.jsx
+++ b/examples/react/js/app.jsx
@@ -20,7 +20,8 @@ var app = app || {};
 		getInitialState: function () {
 			return {
 				nowShowing: app.ALL_TODOS,
-				editing: null
+				editing: null,
+				newTodo: ''
 			};
 		},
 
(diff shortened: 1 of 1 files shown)


samccone:
:+1: this is much better :)

samccone:
@passy looks like the failure is due to one of our child dependencies starting to use es6

The easy fix might be just to bump our travis.yaml to 4
https://github.com/tastejs/todomvc/blob/master/.travis.yml#L8

sindresorhus:
> The easy fix might be just to bump our travis.yaml to node 4

Yeah, no reason not to use the latest Node.js there. I would go with `'stable'` so we don't have to update it for every Node.js version.

samccone:
just a rebase on this @passy and tests should actually run now :penguin:

passy:
Rebased. Now fingers crossed! :v:

PR #1482, opened 2015-10-19
travis update

samccone:
One of our child dependences started to use es6 syntax, because we were locked down to node 0.12 the code actually failed, this should resolve the issue for us

I also dropped an inline comment from our test runner because it was causing an error in the bash script

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@
 sudo: false
 language: node_js
 node_js:
-    - '0.12'
+    - 'stable'
 notifications:
     irc: 'irc.freenode.org#tastejs'
 env:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -5,7 +5,7 @@
 sudo: false
 language: node_js
 node_js:
-    - '0.12'
+    - 'node'
```
sindresorhus: `'node'` → `'stable'`

examples/knockoutjs/index.html
```diff
@@ -1,4 +1,5 @@
 <!doctype html>
+
```
sindresorhus: ?

.travis.yml
samccone: screen shot 2015-10-19 at 9 37 44 am


They mean the same thing 
http://docs.travis-ci.com/user/languages/javascript-with-nodejs/

sindresorhus: I know, but `stable` is clearer.

samccone: k

PR #1484, opened 2015-10-19
Ignore js style linting in typescript examples

samccone:


.jscsrc
```diff
@@ -34,6 +34,8 @@
 		"examples/polymer/elements/elements.build.js",
 		"examples/js_of_ocaml/js/*.js",
 		"examples/typescript-react/js/*.js",
+		"examples/typescript-angular/js/*.js",
+		"examples/typescript-backbone/js/*.js",
```
arthurvr: Wouldn't `"examples/typescript-*/js/*.js",` work?

samccone: I think it might

samccone:
good call @arthurvr 

screen shot 2015-10-19 at 12 50 35 pm

passy:
:+1:

PR #1486, opened 2015-10-19
Include all nested js in glob for typescript

samccone:


samccone:
tested locally :) this one is now working as we expected :smile_cat:

PR #1218, opened 2015-03-19
Convert TroopJS example

mikaelkaron:
Hi.

We've started work on converting and think we have something we can share. Note that this is still preliminary code, but we think it's in a state where we are ready for feedback.

These are know issues:

 - bower used for packaging vs. npm. Hope this is not a problem.

These are outstanding tasks:

- [X] Update "completed count" (pending tastejs/todomvc#1215) [_code is there to update text, hopefylly this will work in the future_]
- [x] Add source comments
- [x] Update readme
- [x] Remove unused vendor files

PR #1491, opened 2015-10-25
Update TroopJS to latest 3.0.x and new template.

mikaelkaron:
Reopen from tastejs/todomvc#1218


mikaelkaron:
Its not clear to me from the travis output what went wrong here @samccone maybe?

samccone:
@mikaelkaron checking out this branch looks like you are missing a few things, I am unable to add new todos (no JS errors), and the styles are missing

screen shot 2015-10-27 at 9 20 56 am

mikaelkaron:
I just did the same, but for me it works fine... let me try to force push, maybe I forgot something?

samccone:
@mikaelkaron  it is very likely there are files that are gitignored that you have locally that are not committed.

To debug this: 

* clone your fork into a clean folder 
* checkout your branch 
* try and view this example, it should match what I see.

mikaelkaron:
There. Should be good to go now.

I guess fiddling with `.gitgnore` is trickier than I expected.

samccone:
@mikaelkaron not quite yet ;)

mikaelkaron:
ohwell. is the UI still broken? or is this about the travis build?

mikaelkaron:
Any pointers to what I should look at, are you able to make a clean pull that looks ok?

samccone:
Hey @mikaelkaron I did some digging and it turns out it was my fault! 
So I ran the test suite and ended up finding some issue around the routing tests.

If you want you can run the tests locally by starting up a local server and then running from the root

`./test-runner`

mikaelkaron:
Ok, I can replicate the routing errors locally with the tests (but I can do that all manually, let me see what is failing).

mikaelkaron:
After a bit of research I'm more confused then when I started... So from what I can see the routing tests actually fail because it can't find the list element by xpath.... I came to this conclusion by first isolating a failing test and executing:

```bash
npm run test -- --framework=troopjs_require --grep 'Routing should allow me to display active items'
```

so we'll fail with this:

```bash
  TodoMVC - troopjs_require
    Routing
      1) should allow me to display active items


  0 passing (6s)
  1 failing

  1) TodoMVC - troopjs_require Routing should allow me to display active items:

      AssertionError: [ 'book a doctors appointment', 'buy some cheese' ] deepEqual []
      + expected - actual

      -[
      -  "book a doctors appointment"
      -  "buy some cheese"
      -]
      +[]
```

however if you manage to spy on the UI when this test runs you can see it actually doing what it's supposed. So I dug into the test and started stepping through `testOps.assertItems` -> `page.getVisibleLabelText` -> `page.getVisibleItemLabels` -> `page.getItemLabels` where I realize that no labels are returned.

So I added some debug code in `page.getItemLabels` to see what's up.

```javascript
var _xpath = this.getTodoListXpath();
this.tryFindByXpath(_xpath).then(function (_result) {
	console.log("tryFindByXpath:getTodoListXpath", _xpath, _result);
});
```

and when I run the test I can see this in the console output:

```bash
tryFindByXpath:getTodoListXpath //ul[@class="todo-list"] []
```

whaaat?

After a bit of digging I think an acceptable solution could be to modify the xpath query returned from `getTodoListXpath` to be `'//ul[contains(@class, "todo-list")]` [here](/tastejs/todomvc/blob/3f4bc5f42099689a2a684df68bb3518beacad3b9/tests/page.js#L19).

I'm not 100% sure why this fixes the problem but the query looks to be in line with other `@class` based xpath queries so I don't think it's a bad change.

If you think this change is OK I can submit a separate PR against master and then rebase.

samccone:
oh wow @mikaelkaron A+++++ research here :clap: thanks SO MUCH

I would love it if you would submit another PR for that tweak (I can also do it but you deserve all of the credit)

I am excited to see this PR finally land, thanks a million for working through it with us :)

samccone:
cool @mikaelkaron if you rebase now.. things should go green :dancer:

mikaelkaron:
Looks like I get no love from travis. I get the feeling the tests start running before the app is "ready" (all js deps are loaded) and thus some elements may not be there, and from that I'm going to go out on a limb and say it has something to do with requirejs.

Is there an "official" way that I can signal the test runner that the app is ready? Maybe something like focusing the input, or starting out with the input disabled and then enabled, really anything that can be observed from the outside (test runner) and that perhaps made some sense in the context of the application (otherwise maybe just triggering a custom dom event)?

mikaelkaron:
To confirm I increased the delay [here](https://github.com/tastejs/todomvc/blob/3f4bc5f42099689a2a684df68bb3518beacad3b9/tests/test.js#L46) from `200` to `2000` and the router tests pass (however the persistence test fails).

samccone:
Hey @mikaelkaron so, obviously this test-runner is generic enough to run across every implementation but each implementation does have some special gotchas... async loading of assets is totally a valid thing, and you are right... it seems like we are not handling it..

So I would take a look at where the first point of failure is, and then we can put some smart retry verify logic inside of a wait (that is generic enough to work on every implementation and will "fix" this case)... and that should take care of things

I wrote up some thoughts about this entire cycle here http://samsaccone.com/posts/fixing-the-flake.html


If you would like to take a stab at this it would be very much appreciated, however I will also take a look at this tonight, thanks for working through this with us :dancer:

mikaelkaron:
If some generic bootstrap can be made that would allow other test runners to be used that would be great (basically if the "detect ready" and the "start test runner" parts were separated and somewhat independent).

I mention this 'cause I've actually talked to someone who uses todomvc as one of his components when teaching test automation and he was looking for exactly this kind of functionality.

If you want some extra hands or viewpoints I'm always available.

samccone:
yeah, "detect ready" is really what we are after here however a simple concept it may be... there are many layers of "ready" and codifying it is the hard (see fun) part

mikaelkaron:
I guess that asking app developers to generate a synthetic event (selecting, clicking or custom) somewhere is not an option? If I remember correctly [the spec already says the "new item" input should be focused](/tastejs/todomvc/blob/master/app-spec.md#new-todo)

> The input element should be focused when the page is loaded preferably using the autofocus input attribute.

If we update that paragraph to say that the element should be focused when the app is "ready" and if it is auto focused we simply hook the browsers 'ready' event.

This would probably mean that we should update async loading examples to defer focus but that feels like fair effort for value. For any existing framework the change is backwards compatible as all we're doing is to wait before each test for the input to receive focus, and there is a test for that already so passing apps should be safe.

samccone:
@mikaelkaron got it all working, a green CI is just a rebase away for you :)
https://github.com/tastejs/todomvc/pull/1499

examples/troopjs_require/js/clear.js
```diff
@@ -0,0 +1,38 @@
...
+	'use strict';
+
+	/**
+	   * Component for the `.clear-completed` button
+	   */
+
+	return Component.extend({
+		/**
+		  * HUB `todos/change` handler (memorized).
+		   * Called whenever the task list is updated
```
samccone: nit: looks like your spacing is off a tiny bit

examples/troopjs_require/.gitignore
```diff
@@ -0,0 +1,39 @@
...
+!node_modules/mu-merge/main.js
+!node_modules/mu-selector-set/main.js
+!node_modules/mu-selector-set/src/*
+!node_modules/poly/lib/*
+!node_modules/poly/support/*
+!node_modules/poly/array.js
+!node_modules/poly/json.js
+!node_modules/poly/string.js
+!node_modules/requirejs/require.js
+!node_modules/todomvc-app-css/index.css
```
arthurvr: This one is a duplicate of line 3

samccone:
#1514 is all :green_apple: 

going to land this now (with #1514 {I took care of the nits for you}), @mikaelkaron thanks for sticking this one out!!!! A+ work all around 

:spaghetti: :school: :dancer: :bee: :beers: :rocket: :space_invader: :heart_eyes_cat: 


photo on 11-9-15 at 10 13 pm

PR #1493, opened 2015-10-27
analyse and learn angularjs with typescript

spatil2014:


samccone:
Hi @spatil2014 since there is no explanation here, I am going to close, let us know if we should reopen.

PR #1498, opened 2015-11-04
Make todo-list XPath more permissive

mikaelkaron:
This patch updates the XPath returned from `page.getTodoListXpath` to test if the `class` attribute contains `todo-list` vs. the previous behaviour where the test was equality. This provides a robust solution that prevents some false positives returned from route related tests in tastejs/todomvc#1491.


samccone:
:+1: great work @mikaelkaron

PR #1505, opened 2015-11-08
Vanilla ES6

lukeed:
Translated the `vanillajs` example into ES6 / ES2015. There are currently no other examples written in ES6.

Semicolons are excluded because of:
1. https://github.com/rse/es6-features
2. http://blog.izs.me/post/2353458699/an-open-letter-to-javascript-leaders-regarding
3. supposed standard for ES6 syntax

Built script is the default output of a Webpack build -- hence the comments and funny spacing. But, after all, the example is showing off the ES6 files!

A suite of dev tools is included for further development.


sindresorhus:
I like the idea of showcasing ES2015, but I think it should just be the existing vanilla JS app upgraded to ES2015 instead of another app.

It should also just use Babel directly as a npm run script. No need to have a complicated build-step. [Example](https://github.com/sindresorhus/pageres/blob/28dd9b3004c05470202f02e0ac0d0ee148a63768/package.json#L17).

passy:
+1 for @sindresorhus' point.

addyosmani:
Imo, it's useful for us to still have both an ES5 variant of the vanilla JS app and an ES2015 version. Beginners unfamiliar with ES2015 could benefit from comparing the 'older' syntax with 'new' and once we hit a point where cross-browser ES2015 support is significantly covered, we can drop the ES5 version completely. 

Keeping the ES5 version around is also useful to us browser vendors from a performance comparison stand-point. We've been wanting to compare ES5 vs. ES2015 vs. Babel output of the same app across TodoMVC and (selfishly) this would be useful to be actively supported. 

Other than that, +1 on the `npm run` script.

sindresorhus:
> Beginners unfamiliar with ES2015 could benefit from comparing the 'older' syntax with 'new'

I don't think this is the place to teach people ES2015. The purpose of the vanilla app is to show the best way to create an app in pure JS instead of using a framework. The best way is with the latest syntax.

> Keeping the ES5 version around is also useful to us browser vendors from a performance comparison stand-point.

That's a valid reason, but nothing prevents us from having the existing app in a different repo for that purpose, which IMHO is different from the purpose of TodoMVC, and more of a useful bi-effect of our comprehensive collection of app.

lukeed:
Thanks for the `npm run` tip, I'll upgrade my branch to rely on just that.

As for the usefulness of this example and/or where it should live, I'll let you guys duke that out since this is your project.

My 2¢ is that TodoMVC is a place to learn and compare frameworks, as well as framework syntax. Since ES2015 is not *yet* the new standard (using browsers' ability to read non-compiled ES2015 as the metric), and with so many "Learn ES6" guides still being pumped out, I think a separate example is totally valid... for the time being.

@sindresorhus is thinking far ahead, when ES2015 just becomes known as JavaScript. At that point, both "vanilla" repos can be merged because there'd no longer be a need or curiosity to compare ES5 against ES6 anymore.

sindresorhus:
> @sindresorhus is thinking far ahead, to when ES2015 just becomes known as JavaScript.

No, he's not, he's thinking now! :) ES2015 is JS, today. Latest Node.js, Chrome, Firefox, Edge already covers the most popular ES2015 features. In my opinion, not knowing ES2015 is not knowing JS.

I don't have strong feeling whether this should be a separate app or not, so I'll leave that up to the other maintainers, but I strongly disagree with the sentiment that people shouldn't be expected know/learn ES2015 today.

lukeed:
@sindresorhus I'm on the same page 100%, lol. But in terms of practicality, ES2015 acts as a separate entity when it comes time to write an app that should be able to run on anything other than the cutting edge releases

ycyxyd:
Hallo,

bitte schmeißen Sie meine Email-Adresse von Ihre List 
raus, ich bekomme jeden Tag unendlich solche Emails von dieser Adresse, 
und nur in Spam wegschmeißen bringt es gar nicht.

vielen Dank

Date: Sun, 8 Nov 2015 10:44:30 -0800

lukeed:
So, to be perfectly honest, I'm not sure what to do about the failing Travis checks...? It's unhappy about formatting & function declarations for the bundled Browserify file

lukeed:
@ycyxyd I did a rough translation, but if you don't want to receive notifications about Pull Requests, you have to update your 'Watch' settings for the TodoMVC repo

samccone:
@lukeed since the files are compiled you will want to add them to the root .jscs config
https://github.com/tastejs/todomvc/blob/master/.jscsrc#L26

lukeed:
@samccone every file that was compiled, or just the final `bundle.js`?

samccone:
@lukeed mind squashing this down into 1 commit for us to review?

lukeed:
fuark

PR #1510, opened 2015-11-09
correct actual vs. expected order in assertions

davidrunger:
See https://nodejs.org/api/assert.html. The "actual" value should come
first, and the "expected" value should come second. This ordering is
important in order for the output describing a failed test to be
correct/accurate.

### DIFF
--- a/tests/testOperations.js
+++ b/tests/testOperations.js
@@ -15,7 +15,7 @@ function TestOperations(page) {
 	}
 
 	function testIsVisible(elements, name) {
-		assert.equal(1, elements.length);
+		assert.equal(elements.length, 1);
 		elements[0].isDisplayed().then(function (isDisplayed) {
 			assert(isDisplayed, 'the ' + name + ' element should be displayed');
 		});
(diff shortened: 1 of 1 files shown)


samccone:
heh this has been bothering me for a while :)

Thx for the fix A+

tests/testOperations.js
```diff
@@ -23,7 +23,7 @@ function TestOperations(page) {
 
 	this.assertFocussedElementId = function (expectedId) {
 		page.getFocussedElementId().then(function (id) {
-			assert.notEqual(-1, id.indexOf(expectedId), 'The focused element did not have the expected id ' + expectedId);
+			assert.notEqual(id.indexOf(expectedId), -1, 'The focused element did not have the expected id ' + expectedId);
```
arthurvr: This assertion should probably just use `id.includes()`, right?

samccone: it could :) ... there is a whole host of improves to the test suite that can be done, any help is *more* than appreciated

arthurvr: Sure, going to take time to clean it up tomorrow.

samccone: :+1: :clap: thx!!!

PR #1512, opened 2015-11-09
angularjs: simplify clearCompleted method

davidrunger:
* Remove unused `completeTodos` variable
* use `Array#filter` instead of `Array#forEach`

### DIFF
--- a/examples/angularjs/js/services/todoStorage.js
+++ b/examples/angularjs/js/services/todoStorage.js
@@ -36,14 +36,8 @@ angular.module('todomvc')
 			clearCompleted: function () {
 				var originalTodos = store.todos.slice(0);
 
-				var completeTodos = [];
-				var incompleteTodos = [];
-				store.todos.forEach(function (todo) {
-					if (todo.completed) {
-						completeTodos.push(todo);
-					} else {
-						incompleteTodos.push(todo);
-					}
+				var incompleteTodos = store.todos.filter(function (todo) {
+					return !todo.completed;
 				});
 
 				angular.copy(incompleteTodos, store.todos);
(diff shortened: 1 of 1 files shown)


samccone:
@davidrunger while you are in this method... would you mind checking the other angular examples as well?

Thx for this improvement btw :smile_cat:

davidrunger:
Sure. I just took a look at the other four Angular examples. Every one of them already has this type of simplified code/logic.

angular-dart:
```
void clearCompleted() {
	items.removeWhere((i) => i.completed);
}
```

angular2:
```js
TodoStore.prototype.removeCompleted = function () {
    this.todos = this.get({ completed: false });
};
```

angularjs-perf:
```js
$scope.clearCompletedTodos = function () {
	$scope.todos = todos = todos.filter(function (val) {
		return !val.completed;
	});
	todoStorage.put(todos);
};
```

angularjs_require:
```js
$scope.clearDoneTodos = function () {
	$scope.todos = todos = todos.filter(function (val) {
		return !val.completed;
	});
};
```

samccone:
sweet :+1:

davidrunger:
FWIW, I missed some Angular examples that don't start with "A". (I didn't realize that there are 7 different Angular implementations!) Luckily, these other examples also do not seem to be in need of attention with respect to their `clearCompleted` methods.

The firebase-angular example uses `forEach` rather than `filter`. With the caveat that I know essentially nothing about Firebase, this seems like the appropriate way to use the Firebase API. We can't do something like `$scope.todos = $scope.todos.filter( ... )` because `$scope.todos` is a special Firebase array that is synced with the Firebase server; we can't simply assign over it with a new instance of a plain JavaScript `Array` that is returned by `filter`.

firebase-angular
```js
$scope.clearCompletedTodos = function () {
	$scope.todos.forEach(function (todo) {
		if (todo.completed) {
			$scope.removeTodo(todo);
		}
	});
};
```

typescript-angular
```js
clearDoneTodos() {
	this.$scope.todos = this.todos = this.todos.filter(todoItem => !todoItem.completed);
}
```

PR #1518, opened 2015-11-11
[backbone_marionette] move backbone.history.start() after controller initialization

okiuim:
Backbone history should be started only after controller is started. Otherwise you cannot navigate directly to states like /#/active


passy:
Sorry, I think I'm misunderstanding the problem here. When I go to http://todomvc.com/examples/backbone/#/active right now, I do get to the right tab. What does the change address?

okiuim:
Hi, this concerns http://todomvc.com/examples/backbone_marionette/#/active

passy:
Ah, I'm such a dummy. Tested and the fix works. Thank you!

okiuim:
glad to help :)

Issue #850, opened 2014-02-15
right-column popup too narrow


tomByrer:
On the homepage, if you mouse over the right-hand column, the modal popups are more narrow.  [Reported before](https://github.com/tastejs/todomvc/issues/767).  Widening browser width does not help.

radoneman:
Please see this fix 
https://github.com/radoneman/todomvc/tree/hotfix-right-column-popup-too-narrow

I'm a first timer :)

arthurvr:
Looks like this is already resolved. (Ref https://github.com/tastejs/todomvc/pull/1519)

PR #1519, opened 2015-11-13
Fix for issue #850

radoneman:
Make columns equal width
Make popover fixed width


radoneman:
Moreover the popover does not show up on Firefox (latest version) but only on Chrome on windows.

samccone:
I think this has been resolved already when @passy updated our home page to the latest polymer version.

Thanks though for the PR @radoneman :)

PR #1521, opened 2015-11-16
Update polymer 1.2.2

ebidel:


ebidel:
We need to change how the td-todos.html does its thing. td-model is the source of truth for the .items property. It's a little convoluted that the items would be passed into the computed props of td-model.html, when it is already the keeper of those elements.

I believe the timing changes effect tests because of changes to iron-selector: https://github.com/PolymerElements/iron-selector/commit/92b0bbc31ef7dbb9bbbd1beb0db75877e3babcca. This removed the .items getter (which returned a live node list) to a end of microtask update of distributed nodes (`.observeNodes`) API.

samccone:
manually merged https://github.com/tastejs/todomvc/commit/28bd694fa507370b1a2867f2ee653f3ed5fa7a0d

passy:
Thanks, @ebidel! :)

PR #1528, opened 2015-11-24
Fixed TypeScript+Angular example check all button functionality.

marcogrcr:
Prevents the user from having to click the check/uncheck button two times.

### DIFF
--- a/examples/typescript-angular/index.html
+++ b/examples/typescript-angular/index.html
@@ -16,7 +16,7 @@
 				</form>
 			</header>
 			<section class="main" ng-show="todos.length" ng-cloak>
-				<input class="toggle-all" type="checkbox" ng-model="allChecked" ng-click="vm.markAll(!allChecked)">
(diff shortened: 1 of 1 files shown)


passy:
@marcogrcr Great catch. Thanks!

@mention-bot I love you.

We'll need to create another ticket to update the compiled artifacts.

samccone:
@marcogrcr Mind submitting the build PR for this?

marcogrcr:
@passy No problem ;)

@samccone What do you mean by build PR?

samccone:
so we vendor the built source files for each project, since you modified the source code of this project you will need to recompile the example and open up a new PR with the updated files.

Let me know if this does not make sense :)

marcogrcr:
@samccone  By compile do you mean running **package.json**'s:

```json
"scripts": {
  "compile": "tsc --sourcemap --out js/Application.js js/_all.ts"
}
```

Since I've modified an ```html``` file, I thought that was it, what am I missing?

samccone:
yep that is what I was talking about, however you are correct since this is only a html file I think we are fine :+1: 

:dancers:

PR #1529, opened 2015-11-24
typescript-angular: Cancel edits using Escape key

dmethvin:
Ref #789 

While building I noticed that the `.js` files generated by the `.ts` files were committed to the repo. When the `npm run compile` command is used, it generates the transpiled `Application.js` from *all* the `.ts` files so these files are not needed and are not refreshed by `npm run compile` anyway. So I removed them in a separate commit.


samccone:
:bow: amazing work @dmethvin 

:+1: from me

examples/typescript-angular/js/Application.js
```diff
@@ -53,6 +53,28 @@ var todos;
 })(todos || (todos = {}));
 /// <reference path='../_all.ts' />
 var todos;
+(function (todos) {
+    'use strict';
+    var ESCAPE_KEY = 27;
```
samccone: nit: can we const here?

dmethvin: Why yes, yes you can. I forgot that TypeScript supported it.

dmethvin: Although the change will be reflected in `TodoEscape.ts` down the diff, this is the compiled file.

examples/typescript-angular/js/directives/TodoEscape.ts
```diff
@@ -0,0 +1,24 @@
...
+	export function todoEscape(): ng.IDirective {
+		return {
+			link: ($scope: ng.IScope, element: JQuery, attributes: any) => {
+				element.bind('keydown', (event) => {
+					if (event.keyCode === ESCAPE_KEY) {
+						$scope.$apply(attributes.todoEscape);
+					}
+				});
+
+				$scope.$on('$destroy', () => { element.unbind('keydown'); });
```
samccone: yay no leaks :)

examples/typescript-angular/js/Application.js
samccone: yep sorry commented on the wrong file :)

samccone:
Landed this PR manually

https://github.com/tastejs/todomvc/commit/7d6eb8964e13a17e8e45260f1c25b506386e7da9
https://github.com/tastejs/todomvc/commit/77201888909aca20d9e7264ff35f7ff22cd82ecd

Thanks a lot @dmethvin :dancer:

PR #1531, opened 2015-11-25
thorax: Cancel editing with the Escape key

dmethvin:
Yeah, me again.

This may fail a bunch of jscs style checks on files I didn't edit, but it's fine for the files I did edit. Let me know if you want me to bring this up to spec for those checks as well.

Changes are pretty straightforward, I changed to `keyup` from `keypress` because browsers never standardized behavior for the latter on important keys like Esc which I kinda need to use here. :smiling_imp:

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -26,16 +26,17 @@
 	"excludeFiles": [
 		"**/node_modules/**",
 		"**/bower_components/**",
-		"examples/vanilladart/**/*.js",
-		"examples/duel/www/**",
-		"examples/duel/src/main/webapp/js/lib/**",
+		"**/generated/**",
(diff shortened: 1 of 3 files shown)


samccone:
oh hey @dmethvin :wave: 

Looks like if you updated our jscs ignore to include `examples/thorax/js/lib/backbone-localstorage.js` this would start to pass :)

Mind doing that? (only because the other tests are gated on passing linting)

dmethvin:
Hay I reorged yer .jscsrc it's really clean now and my OCD is better. Let me know if your sock drawer is messy.

arthurvr:
Neat work! Thanks @dmethvin!

PR #1532, opened 2015-11-27
thorax_lumbar: Cancel editing with the Escape key

dmethvin:
Ref #789

In a separate commit I've fixed a bunch of jshint failures and a style indentation issue unrelated to adding the feature. I can squash these if you prefer.

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -35,6 +35,7 @@
 		"examples/js_of_ocaml/js/*.js",
 		"examples/polymer/elements/elements.build.js",
 		"examples/thorax/js/lib/backbone-localstorage.js",
+		"examples/thorax_lumbar/public/*.js",
 		"examples/typescript-*/js/**/*.js",
 		"examples/vanilladart/**/*.js"
 	],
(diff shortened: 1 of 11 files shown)


dmethvin:
Not sure why my jscs passed and this one did not, I thought I pulled the right .jscsrc file. I can try fixing these as well if you want.

samccone:
seems like we might want to drop base.js from the linting eh :)

dmethvin:
I can't get jscs (2.6.0 installed globally) to work here, it just hangs one core for at least 5 minutes before I give up. It's seeing too many errors so probably the only solution that doesn't involve a bunch of edits is to exclude the whole example from the check.

samccone:
@dmethvin since all the files in public seem to be generated.. how do you feel about just ignoring that entire dir?

dmethvin:
@samccone well i guess i feel pretty good about it.

samccone:
I feel good about it as well :D

samccone:
manually landed, thanks again @dmethvin :wink:

PR #1536, opened 2015-11-29
Update RiotJS example

lukeed:
* Fixed/removed unnecessary method scoping within Riot Tags (`todo` and `todoitem`)
* Switch to the "pre-compiled" approach
    * Better start-up performance
    * More useful **in the wild**
        * Easier to chain / re-use tags (not seen here)
        * Minification & caching applied in more straightforward manner... like any other JS file
* Include `compile` script via the `riot` npm package bin


txchen:
IMHO, compiled style is not a good choice for tastejs, because now we need to checkin both the src and generated file.
And, prefer using `.html` since it gives most editors better syntax highlighting.

samccone:
+1 for a non-compiled approach here, as that is the standard approach with all frameworks that can run in an uncompiled mode.

lukeed:
Non-compiled works fine for an app as small as TodoMVC, but it's surely not ideal or all that practical when working with a full production app.

I'm using Riot in a production SPA. I have well over 60 tags & so using the client-side compiler would yield an additional 60+ script tags, hurting load time & < insert all other reasons for file concatenation >.

Also, running a Chrome Timeline on the non-compiled app, you'll get 3 :triangular_flag_on_post: (red flags) because it's trying to do so much at once. 
With the pre-compiled app, you'll only get 1, which is sourced from todo's `base.js`.

txchen:
as @samccone mentioned, in this todomvc project, uncompiled approach is preferred, for example react is using .jsx. 

I think here we want to show the reader how to write the biz logic in different framework, to help others to choose the framework they like, but not to emphasize on the production best practice. User can still follow the links we provided in example to dig deeper.

I suggest we close this PR.

lukeed:
I'd be fine with that.

But, devil's advocate: TodoMVC is often referred to for comparative performance analysis between the various frameworks, even though that ins't its intended purpose. Here is the [most notable example](https://aerotwist.com/blog/the-cost-of-frameworks/) by Paul Lewis.

In such case, React's size:performance ratio is such shit that no one would/should ever consider using it. However, React users know that this isn't actually its true behavior. But that's the catch -- someone who's already familiar with a given framework knows its true potential more than the window shopper.

Ergo, (imo), TodoMVCs should be brought to a prod-ready or near-prod-ready state. All the source files are available so that *the readers see how to write the biz log in different frameworks*. But this way there's added benefit of (1) showing readers how to go from start to finish and (2) bring all examples to a fair/equal playing field.

samccone:
:+1: we agree @lukeed and are working on ideas how to allow for the "compiled" state while still maintaining ease of readability / editability 

For now we need to be consistent until we figure out the correct guidelines of everyone to follow

As far as the perf analysis, think of it this way... since we enforce these rules on everyone... everyone is hit by the same perf penalty.

lukeed:
Okie doke.

I'd probably go with a required folder structure. Something like

```
index.html
+src
  - app.es6
+build
  - app.js
+dist
  - app.min.js
```

You guys could also automate the `dist` state globally with a NPM hook script task... Minify all `build` scripts with the same settings, then update index.html's script src paths.

(Could very well may just be me, but browsing through all the different examples and seeing files in root, `js`, `scripts`, `generated`, `build`, ..... etc, gets me going.)

samccone:
:wave: @lukeed any thing you need from us to wrap this up?

lukeed:
@samccone nah, if pre-compiled isn't what you're after, then that's fine.

The other changes re: self-vs-this are still valid, but not necessarily PR worthy

samccone:
alright, if you would like to open a PR fixing those items more than welcome

thanks again @lukeed

lukeed:
:ok_hand: No problem. Hit me up when the ideas surrounding production builds have been formalized. Would like to help with that.

PR #1537, opened 2015-12-01
sidebar links text fixes when framework + Require.js

cherouvim:
Hello

In some frameworks (AngularJS, Backbone.js etc) the sidebar shows:

    Example
    Source

    Example
    Demo, Source

But the second "Example" header regards the same framework in conjuction with Require.js. This is consistent across all frameworks which have an example with Require.js. This PR fixes this.

### DIFF
--- a/learn.json
+++ b/learn.json
@@ -1,4 +1,4 @@
-{
+﻿{
 	"angularjs": {
 		"name": "AngularJS",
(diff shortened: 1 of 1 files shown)


samccone:
yo @cherouvim thanks for the PR!

this LGTM :+1:

Issue #795, opened 2014-01-24
App does not update the state of the 'mark all complete' checkbox


ColinEberhardt:
The 'mark all complete' checkbox should be cleared if the all the todo items are checked then the user un-checks an item. i.e. the UI state should remain consistent.

This affects the following:

 - [ ] spine
 - [x] angularjs-perf
 - [x] kendo

See the following test failure:

    $ mocha allTests.js  -R list --framework=spine

      ․ TodoMVC - spine (1/1) No Todos should hide #main and #footer: 215ms
      ․ TodoMVC - spine (1/1) New Todo should allow me to add todo items: 553ms
      ․ TodoMVC - spine (1/1) New Todo should clear text input field when an item is added: 287ms
      ․ TodoMVC - spine (1/1) New Todo should trim text input: 321ms
      ․ TodoMVC - spine (1/1) New Todo should show #main and #footer when items added: 405ms
      ․ TodoMVC - spine (1/1) Mark all as completed should allow me to mark all items as completed: 933ms
      ․ TodoMVC - spine (1/1) Mark all as completed should allow me to clear the completion state of all items: 1025ms
      1) TodoMVC - spine (1/1) Mark all as completed complete all checkbox should update state when items are completed / cleared
      ․ TodoMVC - spine (1/1) Item should allow me to mark items as complete: 868ms
      ․ TodoMVC - spine (1/1) Item should allow me to un-mark items as complete: 826ms
      ․ TodoMVC - spine (1/1) Item should allow me to edit an item: 1098ms
      ․ TodoMVC - spine (1/1) Item should show the remove button on hover: 10ms
      ․ TodoMVC - spine (1/1) Editing should hide other controls when editing: 715ms
      ․ TodoMVC - spine (1/1) Editing should save edits on enter: 1079ms
      ․ TodoMVC - spine (1/1) Editing should save edits on blur: 1168ms
      ․ TodoMVC - spine (1/1) Editing should trim entered text: 1131ms
      ․ TodoMVC - spine (1/1) Editing should remove the item if an empty text string was entered: 1013ms
      2) TodoMVC - spine (1/1) Editing should cancel edits on escape
      ․ TodoMVC - spine (1/1) Counter should display the current number of todo items: 426ms
      ․ TodoMVC - spine (1/1) Clear completed button should display the number of completed items: 870ms
      ․ TodoMVC - spine (1/1) Clear completed button should remove completed items when clicked: 844ms
[...]

PR #1538, opened 2015-12-02
angularjs-perf: Fix the "mark all complete" checkbox

dmethvin:
Ref #795 

Looks like a pretty straightforward logic error, the checkbox appears to work correctly now.


addyosmani:
Lgtm

passy:
Works for me. Thanks for the fix, @dmethvin!

This seems to be a really common bug. @samccone Do you think we could include a test for that?

dmethvin:
> This seems to be a really common bug. @samccone Do you think we could include a test for that?

I wanted to add a unit test but I can't get the test setup to work on my machine. I'm on Windows but not sure that's the problem. The README in the unit test dir seems to make some assumptions about having gulp installed globally before starting and perhaps some other prerequisites that don't apply.

PR #1544, opened 2015-12-04
typescript-backbone: Add routing

### LINKED ISSUE #1487: Add routing to typescript / backbone example
cc @DanielRosenwasser @dprotti  @addyosmani 

Anyone have any interest here, or know anyone that might?

Thanks! :sparkles:

dmethvin:
Fixes #1487 

Seems to be passing unit tests and jscs (thanks to the .jscsrc exclusions on `examples/typescript-*`) but not jshint on the generated .js of course.

I felt dirty for introducing a global for the routing filter criteria but `examples/backbone` does it the same way so ¯\\_(ツ)_/¯.

It really seemed to me that the naming should be `backbone-typescript` to keep all the variants of one MVC framework together in the folder, but I don't know if it's too late for that kind of reorg.

### DIFF
--- a/examples/typescript-backbone/js/app.ts
+++ b/examples/typescript-backbone/js/app.ts
@@ -46,6 +46,8 @@ DEALINGS IN THE SOFTWARE.
 // to persist Backbone models within your browser.
 
 declare var $: any;
+
+// TODO: Use DefinitelyTyped rather than ad-hoc definition here
 declare module Backbone {
 	export class Model {
 		constructor (attr? , opts? );
@@ -56,6 +58,7 @@ declare module Backbone {
 		destroy(): void;
 		bind(ev: string, f: Function, ctx?: any): void;
 		toJSON(): any;
+		trigger(eventName: string, ...args: any[]): any;
 	}
 	export class Collection {
 		constructor (models? , opts? );
@@ -69,6 +72,7 @@ declare module Backbone {
 		last(n: number): any[];
 		filter(f: (elem: any) => any): Collection;
 		without(...values: any[]): Collection;
+		trigger(eventName: string, ...args: any[]): any;
 	}
 	export class View {
 		constructor (options? );
@@ -85,6 +89,16 @@ declare module Backbone {
 
 		static extend: any;
 	}
+	export class Router {
+		constructor (routes?: any );
+		routes: any;
+	}
+	export class History {
+		start(options?: any);
+		navigate(fragment: string, options: any);
+		pushState();
+	}
+	export var history: History;
 }
 declare var _: any;
 declare var Store: any;
@@ -161,7 +175,8 @@ class TodoList extends Backbone.Collection {
 }
 
 // Create our global collection of **Todos**.
-var Todos = new TodoList();
+const Todos = new TodoList();
+var taskFilter;
 
 // Todo Item View
 // --------------
@@ -200,9 +215,10 @@ class TodoView extends Backbone.View {
 		// Cache the template function for a single item.
 		this.template = _.template($('#item-template').html());
 
-		_.bindAll(this, 'render', 'close', 'remove');
+		_.bindAll(this, 'render', 'close', 'remove', 'toggleVisible');
 		this.model.bind('change', this.render);
 		this.model.bind('destroy', this.remove);
+		this.model.bind('visible', this.toggleVisible);
 	}
 
 	// Re-render the contents of the todo item.
@@ -210,7 +226,7 @@ class TodoView extends Backbone.View {
 		this.$el
 			.html(this.template(this.model.toJSON()))
 			.toggleClass('completed', this.model.get('completed'));
-
+		this.toggleVisible();
 		this.input = this.$('.todo-input');
 		return this;
 	}
@@ -218,6 +234,14 @@ class TodoView extends Backbone.View {
 	// Toggle the `completed` state of the model.
 	toggleDone() {
 		this.model.toggle();
+	}
+
+	toggleVisible() {
+		var completed =  this.model.get('completed');
+		var hidden =
+			(taskFilter === 'completed' && !completed) ||
+			(taskFilter === 'active' && completed);
+		this.$el.toggleClass('hidden', hidden);
 	}
 
 	// Switch this view into `'editing'` mode, displaying the input field.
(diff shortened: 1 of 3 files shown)


examples/typescript-backbone/js/app.ts
```diff
@@ -261,6 +284,28 @@ class TodoView extends Backbone.View {
...
+	routes = {
+		'*filter': 'setFilter'
+	};
+
+	constructor () {
+		super();
+		(<any>this)._bindRoutes();
+	}
+
+	setFilter (param) {
```
samccone: typescript supports default params we should use that here :)

examples/typescript-backbone/js/app.ts
```diff
@@ -300,8 +345,13 @@ class AppView extends Backbone.View {
 		Todos.bind('add', this.addOne);
 		Todos.bind('reset', this.addAll);
 		Todos.bind('all', this.render);
-
+		Todos.bind('change:completed', this.filterOne);
+		Todos.bind('filter', this.filter);
 		Todos.fetch();
+
+		// Initialize the router, showing the selected view
+		var todoRouter = new TodoRouter();
```
samccone: const :)

examples/typescript-backbone/js/app.ts
```diff
@@ -300,8 +345,13 @@ class AppView extends Backbone.View {
...
+		Backbone.history.start();
```
samccone: oh hey backbone how I have missed you :wave:

examples/typescript-backbone/js/app.ts
```diff
@@ -319,6 +369,12 @@ class AppView extends Backbone.View {
 				completed: completed,
 				remaining: remaining
 			}));
+
+			this.$('.filters li a')
+				.removeClass('selected')
+				.filter('[href="#/' + (taskFilter || '') + '"]')
```
samccone: :( I hate using the dom like this... but I know it is how other examples work :(

examples/typescript-backbone/js/app.ts
```diff
@@ -339,6 +395,20 @@ class AppView extends Backbone.View {
 		Todos.each(this.addOne);
 	}
 
+	// Filter out completed/remaining tasks
+	filter(criteria) {
```
samccone: missing type :)

examples/typescript-backbone/js/app.ts
```diff
@@ -339,6 +395,20 @@ class AppView extends Backbone.View {
...
+		taskFilter = criteria;
+		this.filterAll();
+	}
+
+	filterOne(todo) {
```
samccone: missing type :walking:

examples/typescript-backbone/js/app.ts
dmethvin: Yeah, there were several things I would do differently including a switch to DefinitelyTyped, but it started to become more of a rewrite and I wasn't up for that. I guess the original author would argue this is a manual form of "DOM diffing" optimization but if so it's a premature one. Even the raw jQuery example doesn't do it, plus it seems like it's not scalable.

dmethvin:
OK, I think all these are fixed now,

examples/typescript-backbone/js/app.ts
```diff
@@ -339,6 +396,20 @@ class AppView extends Backbone.View {
 		Todos.each(this.addOne);
 	}
 
+	// Filter out completed/remaining tasks
+	filter(criteria: string) {
+		taskFilter = criteria;
```
samccone: After thinking about this a bit more, instead of a global var, how would you feel about attaching this "State" to the router and when we need to get the current state of the filter just ask the router via a getter* method?

dmethvin: I think it comes down to what you want to make global. TaskRouter is currently inside AppView and not exposed, plus AppView isn't exposed either. Only the TodoList is exposed via the global Todos and even the AppView listens to events through that. The data is needed in TodoView at the moment. So if you put the current filter in either TaskRouter or AppView you have to expose one or both of them and add a getter for TodoView to call.

I feel like I'm missing something about best practice in Backbone (with or without TypeScript). I usually like to leave the code in a better state than I found it but am not sure if I know enough to do that. If the TodoMVC app is this much of a mess I would not like to see how a "real" app came out!

examples/typescript-backbone/js/app.ts
```diff
@@ -261,6 +285,28 @@ class TodoView extends Backbone.View {
...
+	routes = {
+		'*filter': 'setFilter'
+	};
+
+	constructor () {
+		super();
+		(<any>this)._bindRoutes();
+	}
+
+	setFilter (param: string) {
```
passy: Could you remove the space after the method name to follow the style used in the rest of the file? :lipstick:

examples/typescript-backbone/js/app.ts
passy: > If the TodoMVC app is this much of a mess I would not like to see how a "real" app came out!

Such a great point. There are a lot of complaints about TodoMVC not being complex enough to reflect the requirements of real-world applications, but the test definitely works in this direction ...

examples/typescript-backbone/js/app.ts
dmethvin: Yep, glad to do that. I think the master jscs style recommends the opposite but it can't be enforced on TypeScript. Honestly I prefer no space anyway. :bike: :house:

examples/typescript-backbone/js/app.ts
dmethvin: At least when people ask me which frameworks they should use I can now give them a list that I think they shouldn't use. Learned a lot of that lately. :smiling_imp:

samccone: :+1:

examples/typescript-backbone/js/app.ts
samccone: ^ just this one nit and we can merge :)

dmethvin: Sorry, I forgot to push this. I was on an international flight and didn't have internet at the time. Should be good now.

samccone: :airplane: :fr:

examples/typescript-backbone/js/app.ts
```diff
@@ -261,6 +285,28 @@ class TodoView extends Backbone.View {
...
+	routes = {
+		'*filter': 'setFilter'
+	};
+
+	constructor () {
+		super();
+		(<any>this)._bindRoutes();
+	}
+
+	setFilter(param: string) {
```
samccone: oh can we use a default param here? (just noticed this)

https://github.com/tastejs/todomvc/pull/1544#discussion_r46760969

dmethvin: Yep, sorry i missed it. Should be fixed now.

samccone: no worries, sorry for the nits :goat:

dmethvin:
Anything else needed on this?

samccone:
nope this is great, thanks for dealing with the review cycle on this one :)

Issue #14, opened 2011-09-20
Batman.js example


ktusznio:
It would be great to have Batman.js included here.

PR #1549, opened 2015-12-13
Fixes Long Task Overflow

anubhav7495:
Fixes Long tasks are out of the table in Firefox #14
Because the latest todomvc-app-css file uses classes instead of id's now, updating all the apps to latest file broke their functionality, hence a simple fix in each css file individually.


samccone:
thanks again @anubhav7495 :)

Issue #1553, opened 2015-12-19
Would a non-CSS solution be compatible with the website?


relfor:
Hey,

So I just launched this constraint-based rendering engine, which forgoes CSS positioning (for all block elements), and instead uses HW acceleration. It also has MVC baked into it.

The repository is here: [https://github.com/TheRelforFoundation/LAY.js](https://github.com/TheRelforFoundation/LAY.js)

And here is TodoMVC implemented using it:
(http://codepen.io/relfor/full/VeaVNe/)[http://codepen.io/relfor/full/VeaVNe/]

The entire app consists of 1 JS file, and it doesn't use the `base.css` offered as a part of TodoMVC.
The CSS file is not required as the engine takes care of styling and positioning.
Would you ever accept this engine on your website?

samccone:
I think we would accept it... we would just have to manually audit how things look

arthurvr:
See https://github.com/tastejs/todomvc/pull/1614#issuecomment-247875139 :)

PR #1554, opened 2015-12-20
add LAY

relfor:
Reference: #1553


relfor:
The build showed indentation errors.
Are you'll using jslint?

samccone:
if you refer to the test failures on travis, you will see what you need to fix to get things working,

we have a suite of style, memory, and performance tests :)

PR #1555, opened 2015-12-20
add LAY framework

relfor:


relfor:
closed, problem with squashing commits. will fix and send a new pull req.

PR #1304, opened 2015-05-18
angular2: App Example

samccone:
#### TodoMVC 

Alright, so Angular 2 is in a super early state however I think this demo will be valuable for the community as will the gradual refactor towards the real release of 2.0

This current version has no routing, since the angular2 router is still very hard to get up and working in a typescript friendly workflow.

see:
https://github.com/angular/router/issues/233
https://github.com/angular/router/issues/302

-----

#### Pending Improvements (/help)

There is a nasty bug when you double click on a list item you do not always capture the focus of the input.. I have yet to find a good way to know when the change has been applied and then to manually focus the input .. let me know if anyone has any insight here.

I also would like to find a more ES6 and typescript friendly localstorage library, as compared to the hack that I am doing right now :cry: 

-------

#### Remaining

* Figure out what we want readme.md to say
* Figure out what we want the .json to say

--------

#### Angular 2.0 takeaways 

> Some of this may sound hypercritical however I am trying to just document my experiences 

The current mountain of tooling that is required is going to scare many people for a long time to come, until things become more streamlined and or simpler to get up and running, I fear the community will never take hold. Or rather (the community) will simply have a "boilerplate" gulp file and never really understand how things work, which would be non ideal.

Documentation and examples are still clearly lacking, I can tell that the APIs are still rapidly evolving, however I think that the docs will be the ticket for more people to grab on. Having to go into the source code and dig across outdated blog posts and github issues is not an approachable thing for most people. I would hope that the future work on angular shifts a bit from feature development to documentation and example building. 

Overall I did find (once I was able to get my tools going) working in Angular2 to be much nicer and much more explicit than angular 1.x. The API has matured yet the concepts are similar enough to 1.x that most people should be able to make the mental jump without too much strain.

thanks for the :eyes: :palm_tree:

PR #1557, opened 2015-12-21
Updated to Angular 2 beta

ColinEberhardt:
With Angular 2 reaching beta, I think it's time for the example to be updated.

Here's a quick summary of the changes:

Updated app to Angular 2 beta
  + Removed `@View` decorator
  + Changed directives from snake-case to camel-case
  + Added new dependency to Rx
  + Removed old dependency to traceur (not sure this was actually needed?)

Changes to build process
 + Moved typescript configration to tsconfig.json in order to make it a little easier to understand / visualise

There are also a few things I would like to refactor, but thought it best to save that for a separate PR - this one simply focuses on the changes required for Angular 2 beta.

I'll look at adding routing too.

@samccone I also have a solution to the bug you mentioned in #1304 - you must set focus on the input when it is shown, however, with Angular 2 using zone.js, it is only shown after the VM turn in which your code is executed. This can be fixed with a `setTimeout({ // show the input }, 0)` - although this feels like quite a nasty hack, which is why I am not proposing it as part of the PR!

I also echo your sentiment that "The current mountain of tooling that is required is going to scare many people for a long time to come", I spend far too much time fighting the build / tools, and each time I think I've got the hang of it, something else breaks!


examples/angular2/services/store.ts
```diff
@@ -1,9 +1,5 @@
-/// <reference path="../typings/node-uuid/node-uuid.d.ts" />
-/// <reference path="../typings/storejs/storejs.d.ts" />
-/// <reference path="../typings/angular2/angular2.d.ts" />
-
-import {Injectable} from 'angular2/angular2';
 import * as uuid from 'node-uuid';
+import 'store.js';
```
sindresorhus: What's the purpose of this?

And imports should be extension-less.

sindresorhus:
> Added new dependency to Rx

Why?

examples/angular2/services/store.ts
ColinEberhardt: > store.js exposes a commonjs interface, so why aren't we using that?

My understanding is that the type definitions for store.js define a global variable rather than a module:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/storejs/storejs.d.ts#L21

The import statement `import 'store.js';` is transpiled into a commonjs `require` via the typescript compiler.

To be honest, I'm not sure why this example uses `store.js` rather than `localStorage` directly? However, I was working on the principle of making the minimum changes possible for an Angular 2 beta upgrade.

ColinEberhardt:
>> Added new dependency to Rx
> Why?

Rx is a peer dependency of Angular 2, hence it needs to be explicitly included into any Angular 2 project. This was a recent change (introduced when Angular 2 moved from alpha to beta)

examples/angular2/services/store.ts
sindresorhus: > My understanding is that the type definitions for store.js define a global variable rather than a module:

Sounds like the type definition is wrong or outdated.

https://github.com/marcuswestin/store.js/blob/b115845a60bb0c3922ec1ddfd1d35762cf0fec30/store.js#L12

> To be honest, I'm not sure why this example uses store.js rather than localStorage directly? 

Yeah me neither, I think we should just move to vanilla localStorage, but can be in your followup PR.

samccone:
cc @mprobst @igorminar --- would love to get some thoughts on ^

samccone:
@ColinEberhardt mind splitting this PR into 2 chunks to make review easier (we can squash at the end)... 1 for real changes 1 for compiled / vendor changes

ColinEberhardt:
> @ColinEberhardt mind splitting this PR into 2 chunks to make review easier (we can squash at the end)... 1 for real changes 1 for compiled / vendor changes

Happy to ... but ... not entirely sure how to make such a split. There have been minimal changes to the actual app itself, only those required to make the upgrade to Angular 2 beta. I can't see a way to split into separate commits that are self-container units. Any suggestions?

samccone:
oh it is no big deal, github kinda dies on diffs larget than 20k LOC sometimes.. it is more a convince for some of the external people I am pinging on this

as far as how to split... I was thinking 1 commit changes to the app (no compiled changes like just don't commit / those files).. and then 1 commit for the changed files

Obviously this i no something someone would checkout each commit, however it would just make it easier to review in GH

sindresorhus:
> mind splitting this PR into 2 chunks to make review easier (we can squash at the end)... 1 for real changes 1 for compiled / vendor changes

I would agree on larger PRs, but this one has such a few changes it's easy enough to review as is.

examples/angular2/package.json
```diff
@@ -1,18 +1,19 @@
...
   "devDependencies": {
     "tsd": "^0.6.0",
-    "typescript": "^1.5.0-beta"
+    "typescript": "^1.7.3"
   },
   "dependencies": {
-    "bower-traceur-runtime": "git://github.com/jmcriffey/bower-traceur-runtime#0.0.87",
+    "angular2": "2.0.0-beta.0",
+    "systemjs": "0.19.6",
+    "rxjs": "5.0.0-beta.0",
```
samccone: this peer dep issue seems to be something wrong in angular based on what you said

IgorMinar:
reviewed - some small changes requested + it would be great to sync up the app with the conventions and layout of angular-cli.

the rxjs change is legit, it should be defined as a dependency.

we should also mention somewhere in the docs that Angular team knows about the large payload size. We are working on matching or beating Angular 1 before we go out of beta.

samccone:
word, going to land this as is and we can continue with the next PR :)

ColinEberhardt:
> word, going to land this as is and we can continue with the next PR :)

Awesome - thanks @samccone

PR #1562, opened 2015-12-26
firebase-angular: Correct directive name -- `todoEscape`

dmitriz:
Was called `todoBlur`. 
Making it the 2nd `todoBlur` with none `todoEscape`.

Miraculously worked by getting both listeners on the same element via the `todoBlur` directive, 
and "stealing" the `todo-escape` attribute.
Don't think the "miracle" would happen if both directives were used on different elements.
(Sadly miracles are rare :)


samccone:
oh owch, yeah we were very lucky here

good catch @dmitriz :)

Issue #1560, opened 2015-12-23
Tests miss the lack of focus :)


dmitriz:
This fork has disabled focus on double-click to demonstrate that the tests do not catch the lack of focus:

    npm run test -- --framework=angularjs-perf

https://github.com/dmitriz/todomvc/tree/tests-miss-focus-failure

samccone:
good catch @dmitriz 

https://github.com/tastejs/todomvc/blob/master/app-spec.md#new-todo

> The input element should be focused when the page is loaded

:pencil2:

samccone:
I would have expected this to catch it...

```js
		test.describe('When page is initially opened', function () {
			test.it('should focus on the todo input field', function () {
				testOps.assertFocussedElementId('new-todo');
			});
		});
```
https://github.com/tastejs/todomvc/blob/master/tests/test.js#L91

weird eh

dmitriz:
Doesn't look quite that obvious tho.

https://github.com/tastejs/todomvc/blob/master/tests/test.js#L92

uses

    testOps.assertFocussedElementId('new-todo');

but we don't have the `Id` for the double-clicked element, so looks like changes are needed in several places. Or not?

PR #1564, opened 2015-12-26
Add test to verify correct double click focus

samccone:
R: @passy @ColinEberhardt @dmitriz 

ref #1560


ColinEberhardt:
Nice job, I was thinking of adding this myself. Looks good to me :-)

Is it just the Angular 2 version of the app that fails this check?

samccone:
Oh I did not test all of the frameworks :) just did a simple test with backbone, ember, and angular... and a hacked version of backbone to make sure things fail correctly :)

passy:
Sweet! :ship: away!

PR #1545, opened 2015-12-05
Update todomvc-aurelia 0.2.0 (using Aurelia 1.0.0.beta-1)

mhoyer:
- Fixed bugs
  - #1492 - incorrect ref path in Firefox
  - #1543 - could not be reproduced for me with todomvc-aurelia 0.2.0
- Updated to latest JSPM/NPM dependencies
  - Aurelia 1.0.0-beta.1
  - TodoMVC assets (incl. updated views)

Further information can be found in changelog of original todomvc-aurelia repo:
https://github.com/mhoyer/todomvc-aurelia/blob/release-0.2/README.md#changelog

PR #1567, opened 2015-12-28
Updated JSCS dependency in ./tooling to latest 2.7.0

mhoyer:
... and fixed relative path issue.

As requested by @samccone in #1545.

### DIFF
--- a/tooling/package.json
+++ b/tooling/package.json
@@ -1,9 +1,10 @@
 {
   "private": true,
   "scripts": {
-    "lint": "jscs"
+    "lint": "pushd .. && jscs",
+    "postlint": "popd"
   },
   "dependencies": {
-    "jscs": "^1.13.1"
+    "jscs": "^2.7.0"
   }
 }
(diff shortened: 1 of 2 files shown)


tooling/package.json
```diff
@@ -1,9 +1,9 @@
 {
   "private": true,
   "scripts": {
-    "lint": "jscs"
+    "lint": "cd .. && jscs"
```
samccone: `pushd .. && jscs && popd`

Lets use ^ so we do not mutate directory state... thus making lint re-entrant

mhoyer: Well. This is not possible, as we append additional parameters in ```run.sh```. This would result in such calls:

    pushd .. && jscs && popd "examples/aurelia" "tests/"

;-) But this might work:

    "scripts": {
      "lint": "pushd .. && jscs",
      "postlint": "popd
    }

samccone:
LGTM thanks a million @mhoyer !!

samccone:
oh shucks, looks like CI does not have pushd :crying_cat_face: 
https://travis-ci.org/tastejs/todomvc/builds/99285021#L1419

Will apply a hot fix

PR #1571, opened 2015-12-31
Correcting asynchronous testing

mhoyer:
Investigating the test suite of TodoMVC I found missing asynchronicity of the tests (missing callbacks) and the actual chaining of promises across ```page.js```, ```testOperations.js```, and inside the tests itself. Hence I decided to fix this. All in all I applied these improvements/fixes:

- made all tests (```it()```s) to actually be asynchronous tests by adding and handling a ```done``` callback
- promisified all functions in ```page.js```, ```testOperations.js```, and ```tests.js```:
  - actually ```return```ing the promises
  - chaining the promises with ```.then()```s
  - flattening out some ```.then()``` nestings
- extracted new ```page.waitForVisibleElement()```
- changed ```page.ensureAppIsVisible()``` to wait for ```new-todo``` input box instead of ```todoapp```. Some SPAs are dynamically loading the inner content where it might happen that first tests fail due to timing issues.
- simplified test setup for speed-mode and non-speed-mode
- demystified the ```idSelectors``` variable by inverting its meaning and kicking the general ```!```-negation

### DIFF
--- a/tests/page.js
+++ b/tests/page.js
@@ -175,9 +175,24 @@ module.exports = function Page(browser) {
 		});
 	};
 
+	this.waitForVisibleElement = function (getElementFn, timeout) {
+		var foundVisibleElement;
+		timeout = timeout || 500;
+
+		return browser.wait(function () {
+			foundVisibleElement = getElementFn();
+			return foundVisibleElement.isDisplayed();
+		}, timeout)
+		.then(function () {
+			return foundVisibleElement;
+		})
+		.thenCatch(function (err) {
+			return false;
+		});
+	}
+
 	this.getVisibileLabelIndicies = function () {
 		var self = this;
-		var ret;
 		return this.getItemLabels()
 		.then(function (elms) {
 			return elms.map(function (elm, i) {
@@ -186,18 +201,8 @@ module.exports = function Page(browser) {
 		})
 		.then(function (elms) {
 			return webdriver.promise.filter(elms, function (elmIndex) {
-				return browser.wait(function () {
-					return self.tryGetItemLabelAtIndex(elmIndex).isDisplayed()
-					.then(function (v) {
-						ret = v;
-						return true;
-					})
-					.thenCatch(function () {
-						return false;
-					});
-				}, 5000)
-				.then(function () {
-					return ret;
+				return self.waitForVisibleElement(function () {
+					return self.tryGetItemLabelAtIndex(elmIndex);
 				});
 			});
 		});
--- a/tests/page.js
+++ b/tests/page.js
@@ -1,7 +1,7 @@
 'use strict';
 
 var webdriver = require('selenium-webdriver');
-var idSelectors = false;
+var idSelectors = true;
 
 module.exports = function Page(browser) {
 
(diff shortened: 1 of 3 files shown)


tests/page.js
```diff
@@ -175,9 +175,24 @@ module.exports = function Page(browser) {
 		});
 	};
 
+	this.waitForVisibleElement = function (getElementFn, timeout) {
+		var foundVisibleElement;
+		timeout = timeout || 500;
```
mhoyer: The magic 500 is some kind of a hack at this place. I experienced slow tests when running the clear-complete-button tests against ```backbone``` framework. With Backbone approach the items gets hidden with the filter buttons whereas my Aurelia solution rebuilds the whole items list. Thus, for the backbone variant the ```waitForVisibleElement()``` helper will wait for hidden items to get visible in some cases.
:exclamation: Needs further investigation.

ColinEberhardt:
Thanks @mhoyer - there are some really good improvements here, especially those in `page` and `testOperations`. However there are a number which are not strictly necessary.

The WebDriverJS API has a [promise manager](https://code.google.com/p/selenium/wiki/WebDriverJs#Control_Flows) which removes the need to explicitly chain promises.

Did you actually experience any problems with the runtime functionality of these tests? Or was your aim just to refactor / optimise?

mhoyer:
> However there are a number which are not strictly necessary.
>
> The WebDriverJS API has a promise manager which removes the need to explicitly chain promises.

Interesting. Didn't know this fact. But IMHO having those explicit ```then()```-chains seems less magic to me. Maybe others also don't know about this queueing approach in ```selenium-webdriver```. But I am also fine with removing the unnecessary ```then()```s.

:question: What do you (or others) think? @samccone? 

> Did you actually experience any problems with the runtime functionality of these tests? Or was your aim just to refactor / optimise?

Indeed, I had issues with my Aurelia implementation. E. g. the *"Clear completed button - should remove completed items when clicked"* where Aurelia seems to render the visibility of the clear completed button slightly later as the test runner executes the assertions.

mhoyer:
I added one commit that reverts the unnecessary ```.then()```-chaining in ```test.js```. Alternatively we could introduce the [fat arrow syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) instead to reduce this ```function ()``` noise and to keep the explicit promise handling.

    createStandardItems()
      .then(function () { return testOps.assertItemCount(3); })
      .then(function () { return testOps.assertItemText(0, TODO_ITEM_ONE); })
      .then(function () { return testOps.assertItemText(1, TODO_ITEM_TWO); })
      .then(function () { return testOps.assertItemText(2, TODO_ITEM_THREE); })

vs.

    createStandardItems()
      .then(() => testOps.assertItemCount(3))
      .then(() => testOps.assertItemText(0, TODO_ITEM_ONE))
      .then(() => testOps.assertItemText(1, TODO_ITEM_TWO))
      .then(() => testOps.assertItemText(2, TODO_ITEM_THREE))

ColinEberhardt:
I'm happy to drop the promise 'magic' in favour of a more obvious implementation. The way that webdriver forces you into a world where everything is asynchronous is something I really dislike.

Before merging I would like to run a full set of tests - this will have to sit until Monday when I am back in the office.

Thanks a lot for these updates - really appreciate this.

ColinEberhardt:
OK, so I've had a chance to do some more comprehensive testing, running these updated tests against a range of frameworks and it all looks good to me.

I think I prefer the updated implementation which makes use of the built-in promise manager. 

Basically this gets the :+1: from me.

ColinEberhardt:
Merged via #1573

mhoyer:
:sunglasses:

Issue #1575, opened 2016-01-05
Implementation inconsistencies seem to mislead public bloggers


dmitriz:
[A recent high profile blog](https://medium.com/@housecor/angular-2-versus-react-there-will-be-blood-66595faafd51#.6yd8v3oze) with 700+ likes and 70+ comments is showing this comparison:

>Ember: 580k
Angular 2: 565k (759k with RxJS)
React + Redux: 204k
Angular 1: 145k <-Edit — Had wrong size earlier.

(Angular 1 revised down after I [wrote to the author](https://medium.com/@dmitri145/angular-1-4-8-min-js-is-actually-only-148k-6342b941c9a3#.ieamoacvf).)

Then [I discovered](https://medium.com/@housecor/where-are-you-getting-those-numbers-f76a298ceab#.ctrn11y81), [where these numbers come from](https://aerotwist.com/blog/the-cost-of-frameworks/), which happens to be based on this project, hence the reason for this Issue here.

As appears, the tests there were based on the examples from http://todomvc.com/. Upon inspection, it appears that [the official AngularJS example](http://todomvc.com/examples/angularjs/) is [loading the almost 1MB heavy unminified source](https://github.com/tastejs/todomvc/blob/master/examples/angularjs/index.html#L67), whereas e.g. [the Polymer example](http://todomvc.com/examples/polymer/index.html) is [loading the much smaller `webcomponents-lite.min.js` of only 39k](https://github.com/tastejs/todomvc/blob/master/examples/polymer/index.html#L37)

Whence the drastic difference in the given test results for the two libraries!

I know that performance comparison was not the point of this project but it seems that's how people use it now, unfortunately. :(

addyosmani:
https://github.com/tastejs/todomvc/issues/1541 will address this. I think we should power ahead with enabling this as an option with some of the more often referenced examples.

passy:
Another one Sam found earlier today: https://docs.google.com/document/d/1eDnJ-NpJBjicxVkrZgCwMHewSZUUePVY6LCVPS_swAw/pub

TodoMVC payload sizes and "execution performance" metrics are used here for evaluating the best framework fit.

While it's extremely unfortunate that people use the current state of the project to (mis)inform their decisions, it's also a strong indicator that there is demand for exactly this: baseline framework overhead and startup/runtime performance metrics for a small apps. So :+1: for getting #1541 ready.

samccone:
going to close this issue as being essentially a duplicate of #1541

[...]

Issue #1572, opened 2016-01-03
change angular.js to angular.min.js


anubhav7495:
In the Angular examples, the angular source used in not the minified version and it is 1mb in size. We must change it to use the angular.min.js file instead to improve page load time.

passy:
We don't optimize for load time. This is a learning resource and we want people be able to jump right into the source.

anubhav7495:
I think you misunderstood, i am not talking about minifying the whole source file, only using the angular.min.js instead of angular.js from the npm packages.

samccone:
I think @passy 's point still stands, however @anubhav7495 if you are interested in making a "perf" version of the angular example checkout https://github.com/tastejs/todomvc/issues/1541

passy:
@anubhav7495 In order to fully understand an implementation, you'll most likely check out the library/framework code, too. We want to retain the ability to insert breakpoints and study the execution flow for all apps - unless it's inherently impossible due to transpilation or a similar required build step.

anubhav7495:
oh! i understand. I will take a look at making a perf version. Thanks

PR #1561, opened 2015-12-23
Removed globals, readability, simplified or removed components - replace PR #1559

dmitriz:
This replaces the PR https://github.com/tastejs/todomvc/pull/1559

In addition to it:
- Further simplifications and readability improvements.
- Replaced the "shallow" directive `todo-escape` with the native Angular `ng-keydown`. One less reason to complain Angular is complex :)

Issue #1541, opened 2015-12-02
optional /perf directory for any example proposal 


samccone:
There is a demand on many fronts to enable "production" versions of todomvc apps.

Any app should be able to add an additional /perf directory 

`examples/backbone/perf` 

The folder should contain an index.html file to serve their app in a way that is inline with best practices for said tool/framework.
#### Requirements for perf
- No files will be linted in /perf
- The root examples folder should have a way to build the source to go to /perf (this is the idea case.. however some examples will need a different codebase... if that is the case.. the new src can be placed into the /perf folder for the example)
- Any changes to the app will require the changer to recompile the perf build (we will need to enforce this)
- CI tests will be run on both /index.html and perf/index.html
- learn.json needs a new perf-enabled field per framework
- If an example if perf optimized by default (this is a outlier case) we will place an index.html file inside of perf/ that redirects to the root example

---

@tastejs/todomvc if everyone is OK with this I know there are several people willing to do a pass on a few implementations

addyosmani:
:+1: from me based on comment from https://github.com/tastejs/todomvc/issues/1490#issuecomment-161431962

sindresorhus:
:+1:

passy:
SGTM. What about perf-related changes to the logic, as with angular-perf? Do we continue to keep those separate or is that something you want to cover with this proposal as well?

samccone:
I would like to kill angular-perf and roll it into normal angular

passy:
@samccone So having `perf/` versions of the app that aren't directly created from the non-`perf/` are okay then?

samccone:
hmm good question, i was thinking there would be 1 codebase just compiled differently for perf...

but perhaps we should allow two codebases because i could think of situations where this would be required

passy:
In theeooory, we could have a `perf.patch` that's applied as part of the build, but I'm not sure if that's gonna make things actually easier for anyone.

samccone:
imho... lets just let some people do the perf version and we can see the demands per each request

thoughts?

passy:
@samccone :+1: Let's figure out the details along the way.

samccone:
:dancer:

addyosmani:
> but perhaps we should allow two codebases because i could think of situations where this would be required

[...]

PR #1576, opened 2016-01-06
Angularjs-perf update: change to minified Angular and renaming

dmitriz:
- Switched to minified Angular, urgently needed due to
  https://github.com/tastejs/todomvc/issues/1575, in line with https://github.com/tastejs/todomvc/issues/1541 and https://github.com/tastejs/todomvc/issues/1572
- Changed `TC` to `tC` to address comment https://github.com/tastejs/todomvc/pull/1561#discussion-diff-48415192 by  @IgorMinar
- Update to the most recent Angular 1.4.8


samccone:
imho we should "kill" angular-perf and move this implementation into

`https://github.com/tastejs/todomvc/tree/master/examples/angularjs/perf`

We can just use a meta redirect refresh in this folder so we do not break any old links....


ref #1541

thoughts @tastejs/todomvc ?

passy:
@samccone :+1: It's confusing to have it as a separate item in the menu.

addyosmani:
>  It's confusing to have it as a separate item in the list on the homepage.

:+1: Let's move it.

IgorMinar:
Let's just delete it. With Angular 2 in beta it's not necessary to have
this perf version anymore since A2 is much faster.

addyosmani:
@samccone @passy if we decide to drop the example, here's a PR: https://github.com/tastejs/todomvc/pull/1586

arthurvr:
Can we take another look at this and fix the merge conflicts? :)

dmitriz:
Sure, if there a chance of getting it accepted ;)
Also with A 1.5 out, we can refine it to a more modern component based feel?

I see Angular 1 as playing an important niche role for people who need something better than jQuery but are not prepared to invest in massive tooling required for more advanced frameworks. The "perf" here is also related to the code simplicity the ways I see, perhaps that can be better advertised?

addyosmani:
Friendly ping @dmitriz :)

dmitriz:
@addyosmani I am somewhat puzzled by no comments to my questions, so I am not sure what is expected.

Is there a guarantee that this PR will actually be merged if fixed? And if yes, how soon? 

No comments at all?

And none to this? -
https://github.com/tastejs/todomvc/issues/1566

PR #1585, opened 2016-01-15
Do not exclude es6 example from lint

samccone:
since 
https://github.com/tastejs/todomvc/pull/1567

We can now handle es6 :)


addyosmani:
LGTM. Will wait for Travis to finish running.

samccone:
oh this might cause things to fail IRL but not on the CI run (because it does not detect any changes to this example)

lukeed:
Ah, nice. I had ignored it because I had errors left n' right.

addyosmani:
I'm going to land this as-is and see if there's further work we need to do given it's all green otherwise.

PR #1586, opened 2016-01-15
Drops the Angular performance example

addyosmani:
As requested by @igorminar in https://github.com/tastejs/todomvc/pull/1576#issuecomment-172011581.


samccone:
hmm, I wonder if instead of dropping we should put it inside of the /perf dir inside of the angular 1.x example


People will stilll want to do perf analysis on 1.x vs other things.. and having them actually test against an optimized version would be good right?

addyosmani:
I'm personally fine with us keeping it or dropping it. I can imagine folks wanting to run perf tests against Angular 1.x compared to 2.x.

samccone:
^ this :+1:

IgorMinar:
I'm fine with either

dmitriz:
Not forgetting clearly marking which version is meant for performance testing, to avoid this: https://github.com/tastejs/todomvc/issues/1575

The big problem with the current Angular 1 code is - it looks really old like it was 2 years ago. 
People will look at it and get completely wrong impression that is what Angular 1 is. 

It would really be more helpful (and fair when you compare frameworks) to have something (more) modern like [in this example](https://github.com/tomastrajan/angular-js-es6-testing-example/).

dmitriz:
Then again, following all the last hypes and whistles may look "over-engineered" to other people not fluent in those things... 

So for a beginner, it may look friendlier to also have some "low tech" version - no webpack/browserify, no es6 etc. That would clearly show that Angular can work nicely both ways.

lukeed:
:+1: for keeping somewhere for comparison

addyosmani:
@samccone I'd rather not leave this PR languishing. Under our newer interest in /perf directories, should this just be moved to a perf directory for this particular implementation instead of dropping it?

samccone:
Dropping for now. Thanks @addyosmani

PR #1588, opened 2016-01-21
Vanilla ES6: Use const where possible

passy:
Some places where just overlooked, some seem to indicate that the object is going to be mutated. I still think it's better to use `const` there it prevents accidental reassignment and there's nothing we gain from loosening the strictness here.

Split into a pre-build and post-build commit.

[Review on Reviewable](https://reviewable.io/reviews/tastejs/todomvc/1588)


sindresorhus:
:+1: Totally agree.

samccone:
great stuff @passy

samccone:
manually landed

PR #1591, opened 2016-01-31
Added duet example.

colingourlay:
Live example: http://duet-todomvc.surge.sh/


colingourlay:
I don't expect this PR to be accepted in its current state (even though I believe it is feature complete, based on the guidelines) purely because **[duet](https://github.com/colingourlay/duet)** is an experimental framework, which I haven't documented yet, or really settled on the API.

I'd just like to start a conversation about what is expected of the examples added to TodoMVC, and what I should focus on for **duet** to be considered a candidate.

I maintain another [example app](https://github.com/colingourlay/duet-example-app), which has a README that contains a description of how **duet** works.

addyosmani:
Thanks for your implementation, @colingourlay! We appreciate you going the mile to submit and try to follow our style guide. 

At the moment, TodoMVC is focusing on patches that update existing implementations or new implems for mainstream libraries/frameworks that are not already captured in the existing collection. 

We've had historical challenges keeping apps with new, smaller/unknown for now libraries updated, however if Duet does take off and gets traction we would be happy to reconsider landing it. 

Traction can factor in a few different things: ecosystem around the library, community discussion in the wild, issue tracker activity (more widely used libraries have a hive of activity), stat counters, stars (though these are mostly meaningless today).

colingourlay:
@addyosmani, thanks for taking the time to look at this!

I was more keen to see how the PR process works for something like this, than actually trying to get Duet into the repo. It's more of a lightweight concept than a library (which means it only has a small % of the responsibility in a typical MVC app)y, so I totally understand why it doesn't fit the same mold that the current accepted implementations have.

Through trying to get the Duet example meet the style guide, I did come across a few things in your automated testing that fall over with a lot of libraries that perform more asynchronous operations than your typical MVC framework (in this case, operating across a web worker bridge). This results in a lot more test cases which strictly fail, but would be considered passes from a UX perspective. If I get some time, I'd like to investigate this further, and maybe start a dialog around how we improve the tests without just simply loosening them.

Thanks again. I know how many directions you're being pulled in these days, and I really appreciate that you responded.

PR #1592, opened 2016-02-02
Added angular2-redux example

leejefon:
The existing angular2 example doesn't seem to work anymore.  I added my angular2 version with redux.


ColinEberhardt:
Thanks @leejefon - this looks like an interesting example. However, I'm not sure whether it makes sense in its current form.

There is a PR that introduces an Angular 2 + ES5 example, that differs from the existing Angular 2 + TypeScript example in that it makes a lot more use of Angular components. As discussed in this PR https://github.com/tastejs/todomvc/pull/1581 - my preference is to update the TypeScript example so that it more closely resembles the ES5 one, i.e. by using components.

For that reason, I'd expect any other Angular 2 example to also 'slice up' the application into a similar set of components, otherwise it makes it hard for users to understand the differences.

addyosmani:
> my preference is to update the TypeScript example so that it more closely resembles the ES5 one, i.e. by using components.

I'm in favour of taking this approach too.

PR #1593, opened 2016-02-02
Update todoStorage.js

githubERIK:
I experimented with the code and Deployd backend (http://deployd.com/). It seemed to me that clearCompleted function worked better with my proposed code.

### DIFF
--- a/examples/angularjs/js/services/todoStorage.js
+++ b/examples/angularjs/js/services/todoStorage.js
@@ -37,8 +37,7 @@ angular.module('todomvc')
 				var completeTodos = store.todos.filter(function (todo) {
 					return todo.completed;
 				});
-				
-				angular.forEach(completeTodos, function(value, key) {
+				angular.forEach(completeTodos, function (value, key) {
 					store.delete(value);
 				});
 			},
--- a/examples/angularjs/js/services/todoStorage.js
+++ b/examples/angularjs/js/services/todoStorage.js
@@ -34,18 +34,11 @@ angular.module('todomvc')
 			),
 
 			clearCompleted: function () {
-				var originalTodos = store.todos.slice(0);
-
-				var incompleteTodos = store.todos.filter(function (todo) {
-					return !todo.completed;
-				});
-
-				angular.copy(incompleteTodos, store.todos);
-
-				return store.api.delete(function () {
-					}, function error() {
-						angular.copy(originalTodos, store.todos);
-					});
+                                for (var i = 0; i < store.todos.length; i++) {
+                                    if (store.todos[i].completed === true ){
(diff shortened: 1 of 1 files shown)


samccone:
@githubERIK looks like a few style issues on CI would you mind fixing those up

githubERIK:
Managed to fix those style errors.

passy:
@githubERIK Could you explain why you want to change the code like this?

Also, it seems like you have found a loophole in our style guide, because this still looks very foreign compared to other code in that file and elsewhere. :)

githubERIK:
With the master branch code I got "DELETE (my_localhost_URL_ path) 400 (Bad Request)" 
and the function did not delete records from Delpoyd and did not update the view. I changed the code and it seemed that it started to work. However, I just now noticed it will delete  only one record. Because of that, I now changed the code again and it seems to work well (_It will delete  all completed records_).
All in all, there is a chance that the original code in master branch did not work for me beacuse I might have modified my todoStorage.js http-paths the wrong way.

addyosmani:
Travis still seems to be complaining here :monkey:

githubERIK:
Indeed. Could somebody explain me, what Travis means by the following output?

```
...
/home/travis/build/tastejs/todomvc/tests/node_modules/drool/lib/index.js:27
  return (new webdriver.WebDriver.Logs(driver))
          ^
TypeError: webdriver.WebDriver.Logs is not a function
...
```

githubERIK:
Pull request #1606 fixed Travis `webdriver.WebDriver.Logs` error. The Travis CI build passed.

examples/angularjs/js/services/todoStorage.js
```diff
@@ -34,18 +34,12 @@ angular.module('todomvc')
...
-
-				angular.copy(incompleteTodos, store.todos);
-
-				return store.api.delete(function () {
-					}, function error() {
-						angular.copy(originalTodos, store.todos);
-					});
+				for (var i = 0; i < completeTodos.length; i++) {
+					store.delete(completeTodos[i]);
+				};
```
arthurvr: No need to have a semicolon here. Kinda surprises me that the linter doesn't check for this? :/

examples/angularjs/js/services/todoStorage.js
```diff
@@ -34,18 +34,12 @@ angular.module('todomvc')
...
+					return todo.completed;
 				});
-
-				angular.copy(incompleteTodos, store.todos);
-
-				return store.api.delete(function () {
-					}, function error() {
-						angular.copy(originalTodos, store.todos);
-					});
+				for (var i = 0; i < completeTodos.length; i++) {
```
arthurvr: Just use `forEach`? :)

PR #1597, opened 2016-02-21
add todoMVC example coding with jQuery-UI.

Ahkari:
refer to jquery example . 

In my example , 
All ui widget create form Widgte Constructor . 
All data binding in widget property OPTIONS .

template engine and the router using the same as jquery example.


passy:
Hi,

thanks for the submission. There are quite a few style and functionality
errors, could you please consult our contribution guidelines?

Also, what's the current state of jQuery UI, specifically for SPA
development? Is it still under active development? Who is using it?

Ahkari:
Sorry, I had ignored the contribution guidelines.
Jquery UI can be used in some complex SPA which need large amount custom widgets.
We used it to create a online game editor based on HTML5. 
[China GameEditor-Gamemei](http://zuoyouxi.gamemei.com/qiyun/gamemei_sln_new/gamemei/)
Maybe its not friendly for you because it's language mainly face to Chinese Game Developer.

Im a noob using Github, so I hava no way to solve the build failed problem come from Travis-ci.
So, thanks for response for me.
This pull request maybe delay for a long time...XD

addyosmani:
If we're going off of GitHub releases alone, development appears to have slowed down somewhat: https://github.com/jquery/jquery-ui/releases. I'm on the fence about including this. If we dive into what about JQUI differentiates itself (apart from the UI components), we're probably just looking at the jQuery widget factory. I don't see a huge number of new applications being built with it today, but if there's interest in this app we can continue pursuing it. Suggest we keep this open for a time to determine that before spending too much time evolving it.

dmethvin:
There doesn't seem to be a lot that the UI widget factory adds to the plain jQuery example, at least as far as an MVC pattern goes.

addyosmani:
> There doesn't seem to be a lot that the UI widget factory adds to the plain jQuery example, at least as far as an MVC pattern goes.

I agree with @dmethvin on this in retrospect. Thanks for the contribution @Ahkari! We're going to pass on this example.

PR #1607, opened 2016-03-26
Add React Native

dabit3:
I'm not sure if this belongs here, or if this is out of the scope of what TodoMVC offers as it's a mobile app, but I had it already built in the style of TodoMVC apps so I figured I'd submit it to see if you all were interested in a React Native example.


sindresorhus:
Probably a better fit for our [PropertyCross](https://github.com/tastejs/PropertyCross) project, but might be interesting if you target [react-native-web](https://github.com/necolas/react-native-web) instead of native platforms.

examples/react-native/app/App.js
```diff
@@ -0,0 +1,98 @@
+/**
+ * Sample React Native App
+ * https://github.com/facebook/react-native
+ */
+'use strict';
```
sindresorhus: No need for `'use strict'` in modules.

examples/react-native/app/App.js
```diff
@@ -0,0 +1,98 @@
...
+  View,
+  Text,
+  ScrollView,
+  Component,
+  StyleSheet,
+  TextInput,
+  TouchableHighlight
+} from 'react-native';
+
+let todoIndex = 0
```
sindresorhus: Use semi-colon everywhere.

examples/react-native/app/App.js
```diff
@@ -0,0 +1,98 @@
...
+    let todo = { title: this.state.inputValue, todoIndex: todoIndex, complete: false }
+    todoIndex ++
+    this.state.todos.push(todo)
+    this.setState({ todos: this.state.todos, inputValue: '' })
+  }
+
+  _deleteTodo(todoIndex) {
+    let { todos } = this.state
+    todos = this.state.todos.filter((todo) => {
+    return todo.todoIndex !== todoIndex
```
sindresorhus: Use a inline arrow function.

examples/react-native/app/App.js
```diff
@@ -0,0 +1,98 @@
...
+        </ScrollView>
+        <TabBar type={type} _setType={this._setType.bind(this)} />
+      </View>
+    );
+  }
+}
+
+export default App
+
+let styles = StyleSheet.create({
```
sindresorhus: Use `const` whenever possible.

addyosmani:
@dabit3 Could you work with the PropertyCross project on landing this work (or something similar) upstream there? I think a React Native example would be awesome there but probably doesn't make sense in TodoMVC right now. We could potentially link to this particular implem from the wiki. Thank you once again for all the work on this.

PR #1609, opened 2016-03-29
Typo in app.tsx importing TodoModel from ./TodoModel.

augustjd:
Compiling as downloaded (with `tsc -p js/`) yields the following error:
```
js/app.tsx(12,27): error TS2307: Cannot find module './TodoModel'.
```


remojansen:
:+1:  sounds good to me

PR #1612, opened 2016-04-10
[fix] update node modules in angular2 example

sdtsui:
Angular 2 example does not currently work with the included `node_modules` folder. 
This PR updates the saved modules, allowing the demo to work again.


sindresorhus:
Only include files actually needed for running the app. The rest should be .gitignore'd.

addyosmani:
@sdtsui Would you be interested in updating this to work with the latest stable version of Angular 2?

arthurvr:
It seems like this PR has gone inactive, and we received a PR to update Angular 2 (https://github.com/tastejs/todomvc/pull/1701) so let's handle this there :)

PR #1621, opened 2016-04-22
Request for adding diffHTML to TodoMVC

tbranyen:
Hi TasteJS folks!

This is a PR that contains a snapshot of: https://github.com/tbranyen/todomvc to be included as part of TodoMVC's collection.

This utilizes a tool I've been working on called [diffHTML](https://www.diffhtml.org) that helps write components and applications with declarative *real* HTML and hook into Virtual DOM changes with transition hooks.

It also utilizes Redux to handle all state, since it pairs so nicely with diffHTML it made sense to include over a vanilla approach.

Let me know what you think, thanks for your consideration!

**~Future improvements~**

There are many limitations with the current implementation. I am currently improving diffHTML to accommodate some of the issues listed below:

-  It needs to re-parse the markup on every change, and while this is very performant, it could be pre-generated with a transform (Babel/Browserify/etc).

- Template literals on their own are entirely string-based and very limited. I've started on a [PR](https://github.com/tbranyen/diffhtml/pull/60) to include a tagged-template helper that will make it easier to compose and will help eliminate much of the `addEventListener` noise, by allowing inline binding. 

  Example: `<div onclick=${this.someClickHandler}>`

- Removals are tricky with an entirely string-based diffing algorithm. It's impossible to tell if an element has been mutated or removed if it shared identical characters to an adjacent element. In the future with the above two optimizations we'll be able to work with static vs dynamic parts of the tagged template and know exactly what is removed vs changed.

- Web Animations are awesome, but they are hardly supported anywhere. I'd like to refactor this to use CSS with a more cross-browser-friendly implementation.


tbranyen:
So I guess the `.editorconfig` wasn't optional... I'll fix the source to use tabs and test the linting locally. Stay tuned!

tbranyen:
I don't mind switching the spaces to tabs, but is there a reason why you don't allow trailing commas? I'd rather not change my source code and don't understand that requirement.

samccone:
consistency across code examples. :)

tbranyen:
Fair enough, I think that's a pretty poor reason to artificially limit the language, but I'll comply and open a separate issue to discuss.

tbranyen:
Looks like my example is too fancy for the Web Driver tests. I'll mess with it when I get home and see if I can track down why it's not passing.

tbranyen:
@samccone I just ran the tests locally and I'm seeing the failure:

``` js
  1) TodoMVC - diffhtml New Todo should allow me to add todo items:

      AssertionError: [ 'buy some cheese',
  'buy some cheese',
  'buy some cheese',
  'buy some cheese' ] deepEqual [ 'buy some cheese' ]
      + expected - actual

       [
         "buy some cheese"
      -  "buy some cheese"
      -  "buy some cheese"
      -  "buy some cheese"
       ]
```

I see the browser rendering 4 items, I'm not sure why. Does anything stand out in my code as the culprit? I'll debug more locally too.

samccone:
Looks like the retry logic might be amiss, I will mess around with this, this weekend :)

tbranyen:
I've got most of the tests passing now, but I'm wondering why I have so many in the first place. I based my code off of the angular2 code, seems like that may have been a bad idea, because running:

```sh
npm test -- --framework=angular2
```

has a *ton* of failures.

tbranyen:
Well I'm stumped, I can't figure out why the CI won't represent what I'm seeing locally. I see 7 fails / 23 passes when running the tests.

I tried to bump the deps, but that was a shot in the dark and didn't work.

tbranyen:
@samccone if i block rendering until animations complete in the transition hooks, i see a ton of failures in the tests. my guess is that you're right the retry logic is messed up and the checks don't wait long enough for the DOM to update after the animations.

I didn't realize the router was required now, do you have a preference for what I use?

samccone:
> and the checks don't wait long enough for the DOM to update after the animations.

Yeah this seems to be the real issue :/


> I didn't realize the router was required now, do you have a preference for what I use?
 Hide all checks


Indeed :D no preference about a router, whatever you prefer

tbranyen:
Realized you don't really need a Router at all for this so I just monitor the hash state in Redux and apply filters based off it.

tbranyen:
@samccone would you prefer I squash down all the commits?

tbranyen:
Down to two failures locally:

-   1) TodoMVC - diffhtml Editing should remove the item if an empty text string was entered:
-   2) TodoMVC - diffhtml Routing should respect the back button:

Not sure why either are failing since I can verify they both work.

tbranyen:
And look at that! All tests passing :D I'll squash down these commits and get you a clean PR.

test-runner.sh
```diff
@@ -4,6 +4,7 @@ set -o pipefail
 
 get_changes ()
 {
+	git remote remove current && \
```
tbranyen: I found I needed this in order to run the tests more than once. Is it okay if it lands with this PR?

samccone: 👍

tbranyen:
Awesome, everything has passed and looks good on my end. Ready for review!

:family:

examples/diffhtml/learn.json
```diff
@@ -0,0 +1,2336 @@
+{
```
tbranyen: This was added accidentally, I'll remove it.

test-runner.sh
samccone: soooo it turns out... adding this line causes the diff check that we do (that would run your code) to fail


https://travis-ci.org/tastejs/todomvc/builds/131036325#L448


The fallback behavior in this case is to run backbone ... thus it is passing :)

tbranyen: Welp really? Shit. I suppose that's cheating to just merge this in then. I'll look into getting this working right.

tbranyen:
Well I'm stumped. Locally I've got them all passing except for one:

``` javascript
test.it('should remove the item if an empty text string was entered', function (done) {
  page.editItemAtIndex(1, webdriver.Key.ENTER);

  return testOps.assertItems([TODO_ITEM_ONE, TODO_ITEM_THREE])
    .then(function () { done(); });
});
```

The Backbone example passes this test, but in my local testing there are only two items, and removes the second item.

In order to get the tests to pass locally I need to modify it to:

``` javascript
test.it('should remove the item if an empty text string was entered', function (done) {
  page.editItemAtIndex(1, webdriver.Key.ENTER);

  return testOps.assertItems([TODO_ITEM_ONE])
    .then(function () { done(); });
});
```

Any ideas?

test-runner.sh
tbranyen: Reverted this in the latest push, should see build failures again.

tbranyen:
My current theory as to why the memory test dies immediately is that diffHTML builds up 10k element and attribute objects. They are all optimized to adhere to their shape, so I figure any reasonable engine can generate them fast enough and store them efficiently in memory.

Maybe this takes too long on Travis and it's dying because the page still hasn't yet rendered.

tbranyen:
This is maddening. I've tested locally and I can verify the tests are running under diffHTML. They fail at the first lookup on Travis. However, as we saw with my blunder earlier they work with Backbone.

What is it about my setup that works locally, but fails in Travis where Backbone does not? Grrrr....

samccone:
i will help here @tbranyen give me a few days though.. I am very slammed with work atm.

tbranyen:
@samccone No worries, this isn't super high priority. I'll mess with it this week and ping you next week if I don't make any progress.

tbranyen:
I can reproduce this locally, but only if the server isn't loaded. I'm guessing the `dist/todomvc.js` file isn't being loaded or is erroring for some reason causing the page to not render anything.

I'm doubting the cause being related to load time, especially after bumping the wait to 5 seconds.

My next tests locally will be using xvfb to simulate the Travis env more closely.

addyosmani:
> My next tests locally will be using xvfb to simulate the Travis env more closely.

@tbranyen Thanks once again for working on this. I know it's been a few months since you last looked at the PR. Did you get a chance to play around with xvfb much to see what the cause of failures were?

addyosmani:
Friendly ping :)

addyosmani:
@tbranyen I would really love to land this but can't if there's a reproducible test failure. Do you think you'll get a chance to double check the issue above?

tbranyen:
@addyosmani FYI I've been rewriting/reworking the diffHTML locally while making devtools and the 1.0 release of diffHTML. Things are looking really good so far!

I haven't forgotten and wanna land it too, but version 0.9.1 is nothing like 1.0.0 will be in terms of speed/stability/features. Mind if we keep this open just a tad bit longer?

addyosmani:
Of course! More than happy to wait until you're comfortable with the 1.0 releases and we can then look at updating. Appreciate you still being interested in contributing a diffHTML example :)

tbranyen:
JSCS seems fine now. However, it still seems like issues spawning the browser in Travis :-/ I have to run selenium-standalone and update the selenium URL locally to get the tests to run at all.

```
λ SELENIUM_REMOTE_URL="http://localhost:4444/wd/hub" npm test -- --framework=diffhtml

> todomvc-browser-tests@ test /home/tbranyen/git/tastejs/todomvc/tests
> mocha allTests.js --no-timeouts --reporter spec "--framework=diffhtml"



  TodoMVC - diffhtml
    When page is initially opened
```

tbranyen:
That said, there are nine errors at the moment, I'll work through them and once everything passes perfectly locally I'll try another CI push to see if that helps.

tbranyen:
```
     WebDriverError: unknown error: Element <input class="toggle-all" id="toggle-all" type="checkbox"> is not clickable at point (214, 215). Other element would receive the click: <input onclick="" class="toggle" type="checkbox" false="false">
```

Three errors resulting from this from this. And:

```
  4) TodoMVC - diffhtml Editing should remove the item if an empty text string was entered:
     Error: Expected item text to be book a doctors appointment
Wait timed out after 3001ms
```

tbranyen:
Tried everything I could think of in the amount of time I tried. No go, not sure why the server dies so quickly in Travis. Works great locally.

addyosmani:
Thanks for going full steam ahead here as far as you could get. Are all the tests now passing locally for you? We can verify and land this if there are no other spec issues remaining.

Travis itself has been incredibly flaky for us lately and we've had to rely on local validation when this has happened before.

addyosmani:
@samccone if that's okay with you I should add.

tbranyen:
The only issues left are the `<input onclick="" class="toggle" type="checkbox">` element being found in the position that the `toggle-all` should be in. I can verify that the coordinates produce the wrong area when I use `document.elementFromPoint(214, 215)` so I'm not sure how its supposed to work. I tested using the Backbone example and it does work, so I can at least look into the differences between the two. (They render visually identically in the browser). So might be something else.

tbranyen:
```
> todomvc-browser-tests@ test /home/tbranyen/git/tastejs/todomvc/tests
> mocha allTests.js --no-timeouts --reporter spec "--framework=diffhtml"



  TodoMVC - diffhtml
    When page is initially opened
    No Todos
      ✓ should hide #main and #footer (307ms)
    New Todo
      ✓ should allow me to add todo items (331ms)
      ✓ should clear text input field when an item is added (206ms)
      ✓ should append new items to the bottom of the list (471ms)
      ✓ should trim text input (194ms)
      ✓ should show #main and #footer when items added (218ms)
    Mark all as completed
      ✓ should allow me to mark all items as completed (76ms)
      ✓ should correctly update the complete all checked state (228ms)
      ✓ should allow me to clear the completion state of all items (116ms)
      ✓ complete all checkbox should update state when items are completed / cleared (203ms)
    Item
      ✓ should allow me to mark items as complete (426ms)
      ✓ should allow me to un-mark items as complete (384ms)
    Editing
      ✓ should focus the input
      ✓ should hide other controls when editing (52ms)
      ✓ should save edits on enter (198ms)
      ✓ should save edits on blur (258ms)
      ✓ should trim entered text (206ms)
      ✓ should remove the item if an empty text string was entered (159ms)
      ✓ should cancel edits on escape (183ms)
    Counter
      ✓ should display the current number of todo items (304ms)
    Clear completed button
      ✓ should display the correct text (100ms)
      ✓ should remove completed items when clicked (217ms)
      ✓ should be hidden when there are no items that are completed (270ms)
    Persistence
      ✓ should persist its data (540ms)
    Routing
      ✓ should allow me to display active items (199ms)
      ✓ should respect the back button (343ms)
      ✓ should allow me to display completed items (276ms)
      ✓ should allow me to display all items (371ms)
      ✓ should highlight the currently applied filter (176ms)


  30 passing (28s)
```

Sweeet! I'll do some clean up. I realized that I had to disable transitions due to the tests :-/ Bummer, but it doesn't bring the filesize down a bit and shortens the code.

samccone:
Nice work tim, just let me know when you are ready for me to take a look. Thanks for cranking on this.

tbranyen:
@samccone its ready to be looked at :)

samccone:
Hey tim, looks like your paths are a touch off.

image


Run your server from the root folder of the project, and then visit the page, you should be able to reproduce what I see.

Thanks!

tbranyen:
@samccone I think you may be on an older commit, the latest references a minified version of the bundle:

https://github.com/tbranyen/todomvc-1/blob/0f7f4a23c042f87d0d475c888009406d77724209/examples/diffhtml/index.html#L14

samccone:
steps to break.

```localStorage.removeItem("diffhtml-todos")```

add todo 1
add todo a

// breaks see image
image

tbranyen:
Pretty concerned about the errors you hit. I wasn't able to reproduce them :-/ I was able to reproduce the edit mode after a refresh though.

I'll keep poking around trying to reproduce the `outerHTML` of null. Sounds indicative of an element that was removed or replaced from the page, but not from the component cache... I'll poke around.

PR #1632, opened 2016-05-24
MaidSafe application example

Zatvobor:
application repo: https://github.com/Zatvobor/todomvc

The main purpose of this app is to provide a complete example of MAID application
development routine. This app promotes things like:

-  React, Redux and Redux-Thunk conveniences.
- compile JS and CSS through `electron-compile`.
- testing through `electron-mocha`. Tests are running in `renderer` process.
- using Fetch API to interface a `safe_launcher` (ships with mocks in specs).
- communication to `safe_launcher` over dedicated thread.
- application authorizing with confirmation whether the obtained token is still valid.
- store todo list into safe network.
- package app into a `asar` archive and release as OSX package (PGP signed).

### DIFF
--- a/examples/maidsafe/.babelrc
+++ b/examples/maidsafe/.babelrc
@@ -1,0 +1,11 @@
+{
+  "env": {
+    "development": {
+      "presets": ["es2015", "react"],
+      "sourceMaps": "inline"
+    },
+    "test": {
+      "presets": ["es2015","react"]
+    }
+  }
+}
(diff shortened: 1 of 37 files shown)


addyosmani:
Thanks for sending over this PR!. My two cents: implementations that use non-mainstream stacks provide value in a "this is an interesting approach to the problem space", but are better referenced from our wiki than the main site. We'll get this one added to the wiki during our next sprint.

PR #1634, opened 2016-06-06
Fix #812 Remove routing mark for Olives

### LINKED ISSUE #812: Remove routing mark

podefr:
Hi,

Sorry for the long delay. I finally got to add routing to Olives, which addresses #812.

jshint is passing and so are functional tests, apart from the autofocus test, which actually works when running the app (issue with the spec runner?). Let me know if it needs more love.

Thanks,
Olivier


samccone:
Since /olives/olives-todo.js is compiled, please add it to our jscs ignore
https://github.com/tastejs/todomvc/blob/master/.jscsrc#L26

Thanks

podefr:
@samccone I fixed the linting issues but the build now fails when prepping for the tests:

https://travis-ci.org/tastejs/todomvc/builds/150820828
https://travis-ci.org/tastejs/todomvc/builds/150820162

looks like other builds were failing for the same reason, e.g.:

https://travis-ci.org/tastejs/todomvc/builds/150285315

any clue?

samccone:
checked locally, all is well! thanks for the PR @podefr

PR #1636, opened 2016-06-10
Replace old ember with new

cibernox:


cibernox:
@samccone I think that this PR is still required. The ember version on the web is still the 1.10-beta

samccone:
yeah the deploy stuff is still wonky
https://travis-ci.org/tastejs/todomvc#L458


Looks like it is failing here
https://github.com/tastejs/todomvc/blob/master/test-runner.sh#L17

I am picking away at it :goat:

Issue #1635, opened 2016-06-10
auto deploy is breaking the site...


samccone:
seems to not be copying any of the root files :(

https://github.com/tastejs/todomvc/commit/582130daf509773298fed73cdc8e9f83a8f30fba

samccone:
A+ error gulp A+
https://travis-ci.org/tastejs/todomvc/builds/136589846#L1176

I have bisected this issue down to the ember cli commit... there is a funky file somewhere ---

attemping to bisect further +1 for 🐵 patching gulp src for real logs 🐐

addyosmani:
Thanks for diving in to discover what's breaking the deploys @samccone 🍪 😎

samccone:
With a patch to vinyl-fs lib/src/getContents/index.js we can see the failure is coming from here

![image](https://cloud.githubusercontent.com/assets/883126/16068242/43104e20-3279-11e6-81d2-377aaeb50203.png)

symlink check is failing to detect that this is a folder...

![image](https://cloud.githubusercontent.com/assets/883126/16068258/821bae48-3279-11e6-879c-b1fa4da21969.png)

Prob since we are using an old version of vinyl-fs
![image](https://cloud.githubusercontent.com/assets/883126/16068336/fe284cc6-3279-11e6-8f04-2b17002f17dd.png)


Looks like contra fixed this stuff in 4
![image](https://cloud.githubusercontent.com/assets/883126/16068344/0ef0625a-327a-11e6-80f8-4a7f4b94e8a7.png)
https://github.com/gulpjs/vinyl-fs/pull/57


----

Off to investigate what needs to be bumped for this to work

samccone:
![image](https://cloud.githubusercontent.com/assets/883126/16068407/621dd070-327a-11e6-9c07-9694f7180ed8.png)


gulp is latest ✅ 

gulp depends on old vinyl 
https://github.com/gulpjs/gulp/blob/master/package.json#L41

https://github.com/gulpjs/gulp/issues/1604

🐐

samccone:
Ok a few options

1) figure out how to not symlink during the build phase in ember-cli
2) write or use our own copy task that handles symlinks 

@sindresorhus @addyosmani @passy do y'alll have opinions on this?

samccone:
Here is the origin issue from  Oct 4, 2014 🌵 
https://github.com/gulpjs/vinyl-fs/issues/39

sindresorhus:
[...]

PR #1643, opened 2016-06-25
Exit right away on error.

samccone:
This will prevent invalid deploys when we encounter errors instead of a
partial complete deploy.

ref #1635


samccone:
R: @passy @addyosmani @sindresorhus

sindresorhus:
LGTM

PR #1646, opened 2016-06-30
Improve test suite robustness.

oskargustafsson:
When running the tests on the Todo app for my (not yet submitted) framework I noticed that some test were blinking. The reason for this nondeterminism is that my framework defers DOM updates using requestAnimationFrame, and thus it has not yet updated the DOM when the tests are asserting things.

To fix this I rewrote the test suite back-end, so that it waits for DOM states instead of just testing for them once, at an arbitrary time.

I have run the test suite on all frameworks presented on the TodoMVC front page and there seems to be no regressions. Some framework test suite results of interest:

* KnockoutJS: 2 test cases fail, both before and after my changes.
* Knockback.js: 1 test case fails, both before and after my changes.
* Flight: 2 test cases fail after my changes, 3 before. The current suite reports a false negative, because it does not wait until the main section and footer are hidden.
* TroopJS+RequireJS: Almost all test cases fail after my changes. This is because of a currently uncaught bug in the TroopJS app, where the app eats up input for a while, before it is completely loaded. Repro; type something and hit Enter repeatedly and quickly while the app is loading, see that the text is removed from the input field but no todo item is added.

### DIFF
--- a/tests/testOperations.js
+++ b/tests/testOperations.js
@@ -1,212 +1,100 @@
 'use strict';
-var assert = require('assert');
 
 function TestOperations(page) {
-	// unfortunately webdriver does not have a decent API for determining if an
-	// element exists. The standard approach is to obtain an array of elements
-	// and test that the length is zero. In this case the item is hidden if
-	// it is either not in the DOM, or is in the DOM but not visible.
-	function testIsHidden(elements, name) {
-		if (elements.length === 1) {
-			elements[0].isDisplayed().then(function (isDisplayed) {
-				assert(!isDisplayed, 'the ' + name + ' element should be hidden');
-			});
-		}
-	}
 
-	function testIsVisible(elements, name) {
-		assert.equal(elements.length, 1);
-		return elements[0].isDisplayed()
-			.then(function (isDisplayed) {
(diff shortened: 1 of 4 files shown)


samccone:
I will review this :) thanks soooo much :)

addyosmani:
> I will review this :) thanks soooo much :)

Thanks, Sam! 🍰

tests/testOperations.js
```diff
@@ -1,212 +1,100 @@
...
 
 	this.assertFilterAtIndexIsSelected = function (selectedIndex) {
-		return page.findByXpath(page.getSelectedFilterXPathByIndex(selectedIndex + 1))
-			.then(function (elm) {
-				var failMsg = 'the filter / route at index ' + selectedIndex + ' should have been selected';
-				assert.notEqual(elm, undefined, failMsg);
-			});
+		return page.waitForElement(
+			page.getSelectedFilterCss(selectedIndex),
+			'Expexted the filter / route at index ' + selectedIndex + ' to be selected');
```
samccone: minor typo but no biggie ha

samccone:
this is great thanks so much!

oskargustafsson:
No problem =)

PR #1649, opened 2016-07-07
Updated to react & react-dom 15.2.0 from ancient vers.

mP1:
The current todomvc master uses an ancient version of react.

This is simply update, no new features or bug fixes were made. The only changes to code was updates to match the "new" name of used functions.

I have also grabbed the latest React js files and included them here (15.2.0). I am aware that JsxTrasformer is deprecated and no longer recommended, but to keep this simple and without a trans-pile step, i have left it in.

I have also added the license files for React.


examples/react/js/app.jsx
```diff
@@ -170,7 +170,7 @@ var app = app || {};
 	var model = new app.TodoModel('react-todos');
 
 	function render() {
-		React.render(
+		ReactDOM.render(
 			<TodoApp model={model}/>,
```
mP1: This is the only code change, since this was written the 'render' method was refactored off into ReactDOM.

addyosmani:
Thanks for the PR, @mP1. This is currently causing the build to fail: https://travis-ci.org/tastejs/todomvc/builds/142919960. Could you take a look?

samccone:
A rebase off of master should clean up the majority of issues

samccone:
@mP1 

1 remaining failure

> TodoMVC - react Editing should focus the input:

addyosmani:
Friendly ping @mP1

addyosmani:
Another friendly ping @mP1. We'd ♥️to be able to land this if you can address the one remaining failure Sam mentioned above.

addyosmani:
One more friendly ping :)

addyosmani:
@mP1 one last ping. If you won't have time to update the PR to resolve the below, we might add this directly ourselves and land the commits here as they'd be good to get in. 

> TodoMVC - react Editing should focus the input

FadySamirSadek:
This should be closed as #1767 is using a more recent version of react and is using create-react-app which as @addyosmani  mentioned a lot of beginners are using it so it can replace the current react exmaple

addyosmani:
Thanks, @FadySamirSadek!

PR #1663, opened 2016-08-10
Render the Binding.scala's TodoMVP application into a div instead of body

Atry:
I expect this PR avoids removing learning-bar in http://todomvc.com/examples/binding-scala/


Atry:
Seems not work.
What should I do in order to show the learning-bar?

samccone:
1) add the example here https://github.com/tastejs/todomvc/blob/master/learn.json#L13
2) https://github.com/tastejs/todomvc/blob/master/examples/backbone/index.html#L2 make sure you have a data-framework attr in your html that matches your key in learn.json

That is all :)

Atry:
Thank you!
I created a PR for the learning bar https://github.com/tastejs/todomvc/pull/1664

PR #1665, opened 2016-08-13
Remove unexpected colon

Atry:


Atry:
I also upgraded the Binding.scala version to 8.0.0

PR #1667, opened 2016-08-16
Updated elm example to 0.17.1

halfzebra:
I have updated the Elm example to use ` 0.17.1` version of Elm, preserving all functionality and most of the app design decisions.

## Motivation:
Latest major versions of Elm introduced a lot of improvements, the example based on `0.15.0` is not relevant.

## Changes:
- Removed all the `Signal` wiring, since they are gone with [A Farewell to FRP](http://elm-lang.org/blog/farewell-to-frp)
- Removed port-based routing and added a very basic example, which relies on [elm-lang/navigation](http://package.elm-lang.org/packages/elm-lang/navigation/1.0.0/)
- Removed ports for passing the saved data, now it is done using [Navigation.programWithFlags](http://package.elm-lang.org/packages/elm-lang/navigation/1.0.0/Navigation#programWithFlags)

Focus behavior and local storage interactions are still port-based.

## Possible improvements:
- Remove [elm-community/string-extra](http://package.elm-lang.org/packages/elm-community/string-extra/1.0.2) dependency
- Improve routing, maybe it should rely on a union type like:
```elm
type Route
    = All
    | Completed
    | Active
    
```


samccone:
Thanks @halfzebra I just need to take this for a spin / fix our webdriver code

halfzebra:
Removed focus port, since Elm can trigger focus() on elements now :sunglasses:

samccone:
Checked manually, all is well here.

Thanks for the PR

PR #1670, opened 2016-08-19
Add meteor version of todoMVC

michalsnik:
Hi!

Some time ago there was a Meteor version of TodoMVC, but due to the fact that Meteor dropped their free of charge hosting at `**.meteor.com` the application stopped working and was removed from official todoMVC website. Anyway codebase of the application was pretty old and not up to date.

That situation makes me wrote a new version of todoMVC in Meteor 1.4.x using most recent practices so here I am with PR that removes old code and replaces it with new one.

Because of the fact that Meteor is a full-stack framework it needs a certain environment to operate on, so I'm hosting this on DigitalOcean droplet here http://meteor-todomvc.tk

I added this link to your `index.html` so it point to the right place for a live Demo.

I'm looking forward to hearing from you!

Edit: I fixed JSHint and JSCS issues, but the CI still has some problems, not sure if these are strictly related with my PR, coz I'm pretty sure this code is not causing them.


addyosmani:
> Because of the fact that Meteor is a full-stack framework it needs a certain environment to operate on, so I'm hosting this on DigitalOcean droplet here http://meteor-todomvc.tk

Thanks a lot for working on this and for setting up the droplet. I think our concern with self-hosting or linking to a non-administered version of the app is that if (for example) you move on, it becomes tricky for us to redeploy to the same endpoint. That said, it may be worth linking up to the implementation and deployed version from the wiki for those that want to try it out.

cc @samccone @passy for their takes on whether it's worth trying to land this work again or just link.

michalsnik:
Hi @addyosmani,

I saw that there are already two todoMVC applications that stands on a separate machines, so I though Meteor could too.

Address of an example doesn't necessarily have to be the same for all times in my opinion, especially here where users in most cases are discovering new apps by clicking links on the todoMVC home page.

I'm willing to create another todoMVC (Meteor + React), so I'm looking forward to hearing from you how are we going to handle this :)

addyosmani:
> I fixed JSHint and JSCS issues, but the CI still has some problems, not sure if these are strictly related with my PR, coz I'm pretty sure this code is not causing them.

There have been a few failures on the CI this week that are WebDriver related so it may very well not be your code causing these issues. Let's run locally and check.

addyosmani:
So it looks like you're going to need to rebase this against master at minimum :) We're currently conflicting on .jscsrc.

> Address of an example doesn't necessarily have to be the same for all times in my opinion, especially here where users in most cases are discovering new apps by clicking links on the todoMVC home page.

I'm fine with this being the case given we've made this exception for Meteor in the past. We may need to ping you over time in case the example goes down (while you're happy to maintain it). Is that something you would be down for?

michalsnik:
Hi @addyosmani, It's looks fine now. Thanks for taking care of those CI failures!

I'm happy to maintain this example, so we're good to go :)

michalsnik:
It's been a while since I posted this PR so I updated Meteor to newest version and re-deployed example.

michalsnik:
@addyosmani ☝️

devagrawal09:
Looks like the link is broken... the link redirects to some other website.
Please fix this

FadySamirSadek:
@michalsnik are you still intrested in landing this ?

PR #1672, opened 2016-08-23
git rid of two useless data flow

geekhac:
I think that these two data flow is unnecessary，because the input know these values itself.
and thank you very much ,your code is so pretty that I learned a lot!

### DIFF
--- a/examples/react/js/app.jsx
+++ b/examples/react/js/app.jsx
@@ -20,8 +20,7 @@ var app = app || {};
 		getInitialState: function () {
 			return {
 				nowShowing: app.ALL_TODOS,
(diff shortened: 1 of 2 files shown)


chenglou:
Thanks, but the state is there for a reason. It's the fundamental distinction between an uncontrolled component and a controlled one (more info [here](https://facebook.github.io/react/docs/forms.html#controlled-components); the latter is preferred whenever possible.

There's no duplication of source of truth here, in case that's your worry. The input's text is completely determined by the `value` you pass to it. Essentially you've lifted the source of truth from the DOM into a platform-agnostic state value.

PR #1685, opened 2016-09-01
Add a sugar.js example.

tangbc:
Hi, I have added a new todoMVC example developed by my own repository [`sugar.js`](https://github.com/tangbc/sugar), it's a tiny and easy-using MVVM library.

I build the app with [`sugar.js`](https://github.com/tangbc/sugar) as MV* framework, [`director.js`](https://github.com/flatiron/director) as router controller and [todomvc-app-template](https://github.com/tastejs/todomvc-app-template) as template structure, it was easy to implement and work well.

I hope this addition can be accept, thanks :-)


tangbc:
I'm sorry I have created two pull requests, first I didn't notice the `jscs` check, secound (this pull request) I fixed all of it.

addyosmani:
Thanks for your implementation, @tangbc! We appreciate you going the mile to submit and try to follow our style guide. 

At the moment, TodoMVC is focusing on patches that update existing implementations or new implems for mainstream libraries/frameworks that are not already captured in the existing collection. We've had historical challenges keeping apps with new, smaller/unknown for now libraries updated, however if this does take off and gets traction we would be happy to reconsider landing it.

Thanks again!

tangbc:
Thank for any of reply, I've seen that there're already exist the same mode mvvm framework implementations, they are very excellent and popular, I still need to work harder ! 

Thank you all the same! I learned a lot from TodoMVC, thank you!

 :-)

PR #1686, opened 2016-09-01
Removed 'bower_components' folder (good practice).

felipexw:
It's a good practice not putting dependencies inside a repo.
Besides, the the packege.json file already contains all dependencies.


samccone:
We do this for stability when deploying and testing locally. 
We remove an entire vector of flake.

felipexw:
@samccone  hmmm, I see.
Sorry for bothering.

PR #1689, opened 2016-09-09
ngBackbone example

dsheiko:
> [ngBackbone](https://github.com/dsheiko/ng-backbone) is a small extension of Backbone.js written in TypeScript that unlocks Angular-like programming experience

Backbone provides us just a minimal abstraction layer and it does quite well. It is also a ridiculously small framework However comparing to the other popular frameworks such as Angular, React, Vue, with Backbone it feels like we stuck in the Middle Ages. We have no data binding, no clean way to handle subviews, working with forms can be a headache and so on.

What I've personally been missing was a few modules on top of Backbone, that would unlock me on true modern development experience. Without much extra code, super easy to learn, but drastically reducing the amount of code required to solve a task. I believe this is exactly what the [ngBackbone](https://github.com/dsheiko/ng-backbone)  offers


dsheiko:
Well, the source code is TypeScript (`app` directory), but CI runs the code sniffer  `jscs "examples/ng-backbone" "tests/"` on the compiled JavaScript. Can somehow to point linter to TypeScript instead of generated JavaScript?

addyosmani:
Thanks for sending over this PR. My two cents: implementations that use non-mainstream stacks provide value in a "this is an interesting approach to the problem space", but are better referenced from our wiki than the main site. 

cc @tastejs/todomvc in case anyone has a different take on this.

sindresorhus:
> My two cents: implementations that use non-mainstream stacks provide value in a "this is an interesting approach to the problem space", but are better referenced from our wiki than the main site.

👍

addyosmani:
With Sindre's +1 on this, I'm going to go ahead and close this PR. We again appreciate your contribution! We'll include a link to this in the wiki.

dsheiko:
Thank you guys!

Yours sincerely,
Dmitry Sheiko
*Senior web developer at CRYTEK GmbH*


*Mobile  +49 17 / 61 256 25 70www.dsheiko.com
<http://www.dsheiko.com> skype: dsheiko*

PR #1701, opened 2016-10-08
Update the version of angular2 to the final release

SangKa:
The Angular 2 example is still use the version of beta, it's out of dated. 
The purpose of this pull request is to update the version to the final release.


examples/angular2/systemjs.config.js
```diff
@@ -0,0 +1,40 @@
...
+      '@angular/common': 'npm:@angular/common/bundles/common.umd.js',
+      '@angular/compiler': 'npm:@angular/compiler/bundles/compiler.umd.js',
+      '@angular/platform-browser': 'npm:@angular/platform-browser/bundles/platform-browser.umd.js',
+      '@angular/platform-browser-dynamic': 'npm:@angular/platform-browser-dynamic/bundles/platform-browser-dynamic.umd.js',
+      '@angular/http': 'npm:@angular/http/bundles/http.umd.js',
+      '@angular/router': 'npm:@angular/router/bundles/router.umd.js',
+      '@angular/forms': 'npm:@angular/forms/bundles/forms.umd.js',
+
+      // other libraries
+      'rxjs':                      'npm:rxjs'
```
reyronald: Is this example using rxjs for anything at all?

examples/angular2/tsconfig.json
```diff
@@ -1,15 +1,13 @@
...
     "experimentalDecorators": true,
     "removeComments": false,
-    "noImplicitAny": true
-  },
-  "files": [
-    "app/bootstrap.ts"
-  ]
+    "noImplicitAny": true,
+    "suppressImplicitAnyIndexErrors": true
+  }
```
reyronald: Can we specify a different --outDir so that the transpiled JS files don't mix with the TS files? I think it will be easier for everyone to inspect the source code that way. What do you guys think?

reyronald:
I agree with @SangKa that this example should be updated! Thanks for the PR.

I submitted some comments and would like to know what you think before this is merged.

examples/angular2/systemjs.config.js
SangKa: rxjs not for this example at all, you can remove it

examples/angular2/tsconfig.json
SangKa: For simplicity there's no need for do this, the TS files for the developer and the JS files for the browser. Do any developer will care the output JS files?

examples/angular2/package.json
```diff
@@ -1,16 +1,32 @@
 {
   "private": true,
   "scripts": {
-    "dev": "tsc --watch"
-  },
-  "devDependencies": {
-    "typescript": "1.7.5"
+    "start": "tsc && concurrently \"tsc -w\" \"lite-server\" ",
+    "typings": "typings"
```
SangKa: I agree

examples/angular2/tsconfig.json
reyronald: When inspecting the example source code all a developer should care about is the TS files, since the JS are generated by the compiler. Having them mixed is unnecessary noise in my opinion, and could potentially throw away a newbie developer maybe?

This is the approach taken for actual projects built with any tech stack that involves some sort of transpiling, so it would be beneficial for this example to reflect that.

SangKa:
@reyronald  thanks for your pretty nice suggestion and I have benefited from this.

reyronald:
Something else:

The Angular team suggests that we add the `@Injectable` decorator to every service regardless, here's the reason:

> We recommend adding @Injectable() to every service class, even those that don't have dependencies and, therefore, do not technically require it. Here's why:
* Future proofing: No need to remember @Injectable() when we add a dependency later.
* Consistency: All services follow the same rules, and we don't have to wonder why a decorator is missing.

[Source](https://angular.io/docs/ts/latest/guide/dependency-injection.html#!#injectable)

I suggest we add the decorator to the `TodoStore` based on that.

Also, we could simplify the `TodoApp` component by leveraging TypeScript's shortcut syntax for constructor parameters.

```typescript
 export default class TodoApp {
 	newTodoText = '';
 
 	constructor(private todoStore: TodoStore) {
 	}
        // ....
```

examples/angular2/package.json
```diff
@@ -2,22 +2,22 @@
...
+    "@angular/forms": "~2.1.0",
+    "@angular/http": "~2.1.0",
+    "@angular/platform-browser": "~2.1.0",
+    "@angular/platform-browser-dynamic": "~2.1.0",
     "@angular/router": "~3.0.2",
 
     "systemjs": "0.19.39",
     "core-js": "^2.4.1",
     "reflect-metadata": "^0.1.8",
-    "rxjs": "5.0.0-beta.12",
```
reyronald: Also remove it from the `system.config.js`

SangKa:
hi @reyronald . About the @Injectable, it is a sort of the best practice. For the respect to the author of this example, I try to keep the code modification as less as possible unless updating the breaking change.  I will left these optimization stuff for the original author. If he doesn't want to do these, then I will try.

blacksonic:
Build fails just as in the ES2015 example https://github.com/tastejs/todomvc/pull/1581

I would add that @Injectable decorator to TodoStore, because it is a recommendation by the core team.

addyosmani:
It would be great if we could get a person or two from the Angular team to sign off on this implementation following best practices. cc @RobWormald @IgorMinar @alxhub in case you have time.

examples/angular2/index.html
```diff
@@ -5,10 +5,18 @@
 		<title>Angular2 • TodoMVC</title>
 		<link rel="stylesheet" href="node_modules/todomvc-common/base.css">
 		<link rel="stylesheet" href="node_modules/todomvc-app-css/index.css">
-		<script src="node_modules/angular2/bundles/angular2-polyfills.js"></script>
-		<script src="node_modules/systemjs/dist/system.src.js"></script>
-		<script src="node_modules/rxjs/bundles/Rx.js"></script>
-		<script src="node_modules/angular2/bundles/angular2.dev.js"></script>
+		
+		<!-- Polyfill(s) for older browsers -->
```
addyosmani: Why is the indentation for this comment inconsistent with the next line where you ref the shims?

examples/angular2/package.json
addyosmani: 👍 For the `postinstall`

examples/angular2/package.json
```diff
@@ -1,16 +1,32 @@
...
   "dependencies": {
-    "angular2": "2.0.0-beta.0",
-    "systemjs": "0.19.6",
-    "rxjs": "5.0.0-beta.0",
+    "@angular/common": "~2.0.2",
+    "@angular/compiler": "~2.0.2",
+    "@angular/core": "~2.0.2",
+    "@angular/forms": "~2.0.2",
+    "@angular/http": "~2.0.2",
+    "@angular/platform-browser": "~2.0.2",
```
addyosmani: Fyi, a lot of these have been bumped to 2.2.3 of late. Could we test this PR is definitely working with the latest?

examples/angular2/package.json
```diff
@@ -1,16 +1,32 @@
...
+    "@angular/platform-browser-dynamic": "~2.0.2",
+    "@angular/router": "~3.0.2",
+
+    "systemjs": "0.19.39",
+    "core-js": "^2.4.1",
+    "reflect-metadata": "^0.1.8",
+    "rxjs": "5.0.0-beta.12",
+    "zone.js": "^0.6.25",
```
addyosmani: rxjs has been bumped to 5.0.0-rc.4. Can we test with that version or does the latest angular 2 not work with the RCs?

examples/angular2/readme.md
```diff
@@ -22,10 +22,11 @@
 
 ## Implementation
 
-This app was built using TypeScript and Angular 2 beta. To make changes simply
+This app was built using TypeScript and Angular 2 final release. To make changes simply
```
addyosmani: To make changes, run the following commands in your terminal:

examples/angular2/package.json
addyosmani: Per comment in another commit, we should be able to drop rxjs as a dep in the package.json entirely.

addyosmani:
Friendly ping @SangKa

PR #1709, opened 2016-10-26
Point to Elm Slack instead of IRC

evancz:
Once someone in the community created a Slack channel, pretty much all conversation migrated over there. These days it is a great place to ask questions and have a friendly discussion about Elm, so I think this would be a more helpful link for people.

### DIFF
--- a/learn.json
+++ b/learn.json
@@ -349,8 +349,8 @@
 				"name": "Libraries",
 				"url": "http://elm-lang.org/Libraries.elm"
 			}, {
-				"name": "#elm IRC Channel on Freenode",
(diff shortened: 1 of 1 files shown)


arthurvr:
Thanks! The IRC channel still has ~200 members and still seems active so maybe we should just mention both the IRC channel and the Slack one? I personally don't like slack to be honest, and I'm sure I'm not the only one.

arthurvr:
I lurked on the IRC channel for a few days and yeah, it's not very active anymore. Merging! Thanks for the PR :)

PR #1710, opened 2016-10-26
Update Elm links in learn.json to point to latest URLs

evancz:
The old links are currently redirected to the new links. I figured I'd skip the redirect while I was looking at the file!

It may also be worth linking to [The Official Guide](http://guide.elm-lang.org/) in addition to these links. Not sure!

### DIFF
--- a/learn.json
+++ b/learn.json
@@ -341,13 +341,13 @@
 				"url": "http://elm-lang.org/"
 			}, {
 				"name": "Guides",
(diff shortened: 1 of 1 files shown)


passy:
@evancz Thank you for that! And yes, I think linking to the official guide would be a good idea.

passy:
I've added a link here: https://github.com/tastejs/todomvc/commit/9f2214d25b4089ef50889626a17656276fa70d3e

PR #1712, opened 2016-11-03
Addition of an example using Nucleon JS framework

moduleon:


arthurvr:
Nucleon doesn't have much of a userbase yet so I'm -1 on adding it right now, sorry. The project needs some time to mature first. I'm also very hesitant to promote a framework that doesn't have any tests, or did I just overlook those? :)

Issue #1648, opened 2016-07-06
React in ES6?


NullVoxPopuli:
it looks like the react stuff is ES5, which adds a lot of prototype bloat. :-(

Hum4n01d:
+1

dhruvdutt:
#1752

FadySamirSadek:
Closed as this is now already implemented.

PR #1721, opened 2016-11-24
React in ES6

zipengwu:
This PR implements #1648 and it updates React and React-Router to latest versions


zipengwu:
Hi, all tests passed on my machine with node 6, but it failed on travis with not much info. can someone have a look ?

addyosmani:
@arthurvr Noticed you tagged the original issue. Are you thinking there's value in an ES6 specific version of the React implementation?

@zipengwu Thanks for the PR. I've restarted the build. We're running into Travis breaking before testing folks changes this week due to a WebDriver error. Looking into it.

arthurvr:
@addyosmani Ah, yeah, I gave all requests for new apps that `new app` label. Maybe `new app request` or something would've been more appropriate.

> Noticed you tagged the original issue. Are you thinking there's value in an ES6 specific version of the React implementation?

I don't think it's worth it to have separate versions of the example apps. Lots of effort would go into maintaining two apps which are really the same framework, and React sure isn't the only app that would come up. 

I'm all 👍  for updating the existing apps to ES6 as soon as it's browser-support-wise possible, but I don't really think we're at that point yet. WDYT?

Nice work though, @zipengwu! Thanks for working on this. It might not be directly right now but it will soon proof useful in the future :)

zipengwu:
Hi @arthurvr, the app is written in ES6 and transpile into ES5 for browser support, and this is a common practice these days. If you look into the code, you will see the project uses official [react cli](https://facebook.github.io/react/blog/2016/07/22/create-apps-with-no-configuration.html) to avoid boilerplate.

passy:
I do agree that having two examples that should ultimately result in equivalent code doesn't seem worth the maintenance overhead. However, as @zipengwu says, especially for React applications it's very common to use a transpilation step these days and one could argue that it's less idiomatic to stick with ES5.

I'm not sure what the best solution for us is - perhaps someone working on React could chime in and give some input on what they'd like to see represented here.

FadySamirSadek:
Thank you, but I am going to close this as a similar, more recent PR has been merged

PR #1723, opened 2016-12-09
Refactor of the Typescript-AngularJS TodoMVC

jmitchell38488:
This is a significant refactor to change how the angular relationship with the DOM works. Instead of using controllers and explicitly defining them, I have replaced them with directives that are attached to the DOM using `<todos-app></todos-app>`, as well as replace `vm.whatever` with the controller `todosApp.whatever`.

Full list of changes:
 - Updated angular to 1.6.0
 - Updated TypeScript to 2.1.4
 - Added angular-route dependency
 - Removed all instances of `'use strict;'` since this can be controlled with the TypeScript compiler
 - Replaced all instances of `module todos` with `namespace todos` since they fundamentally mean different things and `module` is incompatible with NodeJS
 - Removed manual injection in the TodoController
 - Removed 'controllers' directory
 - Created sub-directories in the `directives` directory for each directive, moved existing directives into their respective folders
 - Created `directive` sub-directory for `todosApp`. This is the refactored Todo Controller
 - Renamed the controller name to `TodosAppController`
 - Added `directive` and `controller` definition to the module inside the defining namespace instead of the root module, since this solves several loading issues, also allows greater control of what is and isn't available
 - added `: any` to the filterFilter property in the `TodosAppController` constructor
 - Cut the template portion related to the former todo controller and moved it into the directives folder, naming it template, `directives/todosApp/template.html`
 - Added `<todos-app></todos-app>` as the placeholder for the todosApp directive
 - For each module extension that was originally defined in Application.ts, I have moved their module definitions into their respective files. This allows us to inject `module.ts` first and make it available to everything, so that angularjs injection detection can happen automatically without explicitly defining what's to be injected in each file (`$inject['blah'`)
 - Renamed `ITodoStorage` and service to `ITodoStorageService` so that it is clear what the purpose of it is

It would be nice to add unit tests and e2e tests. For another time I think. I also added my details as a contributor for this package.


jmitchell38488:
Anyone reviewing this?

addyosmani:
Thanks for taking the time to work on this refactor. We haven't landed too many changes to it in a while. I'm happy to review your PR but wonder if there are any other TS/ng experts in the community worth looping in just to gut check the changes otherwise still reflect best practice. Any thoughts on folks that might fit that bill?

addyosmani:
Glancing over the source, these changes look fine to me for the most part. When you suggest 'namespace' is incompatible with Node, could you expand on what you mean?

PR #1746, opened 2017-01-26
Renamed indexFromEl function to getIndexFromEl

sindreon:
All other functions on the App object are in verb form. This change
ensures all methods have consistent naming form.


passy:
Thanks for the improvement, @sindreon!

PR #1747, opened 2017-01-27
Renamed edit function to editingMode

naterino:
The edit function actually switches the todo to an editable input. It doesn't edit anything itself. The new function name 'editingMode' is more accurate and descriptive.


passy:
Agreed. Thanks for the pull request! :)

PR #1749, opened 2017-01-28
angularjs: do not use self-closing form of <ng-view /> tag

### LINKED ISSUE #1748: AngularJS example is broken with Angular 1.5.0+
This prevents using the example with AngularJS 1.5.x and higher.

As explained in angular/angular.js#15642, the `<ng-view />` form must not be used anymore. The one that works is `<ng-view></ng-view>`.

kveretennicov:
Fixes #1748 by using full-form `<ng-view></ng-view>`.


passy:
Thanks for the fix, @kveretennicov!

PR #1750, opened 2017-01-30
Input field on edit wasn't getting focus

robnagler:
In order to get focus on the editable input, added FocusDirective (todo-focus), which calls focus on the input. The only way to get this to work was to use setTimeout, which is clunky, unfortunately.


robnagler:
Sorry, I forgot setTimeout() messes up protractor. I'll look into using [Observable](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#static-method-timer).

robnagler:
The master is [failing](https://travis-ci.org/robnagler/todomvc/builds/196461734) for the input test and routing tests so I think I improved things. LMK.

PR #1757, opened 2017-02-23
added typescript-react-tagged

Baavgai:
For your consideration, my implementation of todomvc using newer TypeScript features and a significantly different design than the other typescript-react.  Any feedback appreciated.


Baavgai:
After addressing a single pedantic lint rule on a config file ( I do NOT agree with tabs ) the test appears to fail on its own.  At least it's consistent: I got the same results when I tried to run on my box.  Figured I'd just have to go with the Travis thing.

PR #1765, opened 2017-03-11
hyperHTML does TodoMVC

WebReflection:
This [new approach](https://medium.com/@WebReflection/hyperhtml-a-virtual-dom-alternative-279db455ee0e#.9hbn76rwm) to map template literals to DOM nodes
and update only target nodes, either attributes, text,
or even HTML and fragments, can shows its muscles and
potentials through this application too.

### Why hyperHTML should be included?

Lurking a bit through various examples, [hyperHTML](https://github.com/WebReflection/hyperHTML) seems to win in terms of readability, ease of use, performance, and production final size, requiring less than 2KB to run transpiled and browserified code, and with a library that so far weights about 1.5KB.

As example, the test suite run on my laptop in a total of 42 seconds, compared to 43 seconds spent for Vanilla JS test.

Please consider adding `hyperHTML` to the list of viable alternatives to common frameworks and libraries, and don't hesitate to let me know if I should update or change or do anythin else for this PR to succeed.

**Thank you**!


WebReflection:
OK I don't know what is going on in Travis-CI

I have successfully run `./run.sh --framework=hyperhtml` in both `tooling` and `tests` and everything is green and nice.

Please let me know what I am doing wrong, thanks.

PR #1767, opened 2017-03-19
ES6 create-react-app TodoMVC

jchris:
Facebooks new `create-react-app` tool provides a nice running and live code editing etc. I've taken the existing default React TodoMVC and hoisted it into the upgraded environment. I made the smallest changes I could to the code base to get it to run with the latest react.

This change should be a non-controversial simple upgrade, making the outdated React TodoMVC modern. There is still room for the code to be changed more to take advantage of es6 transpiler features, but I think just getting it onto the new runtime is the most important part.


jchris:
It looks like the tests are complaining about syntax stuff, so maybe I just need to adjust an .eslint file or something?

addyosmani:
The tests have been flakey for a while (our apologies, we're looking into it). From what I observe, there are two changes being introduced here:

- Updates to latest React
- Rewrites the existing React implementation using CRA

From my perspective, a lot of beginners are going to use CRA rather than what we have in place here at present so I'm fine with moving ahead with that side of these changes. 

@jchris Next steps here would be to loop in some folks from CRA to validate the implementation holds up to their best practices. Would you like to nominate one or two folks that might be able to do that from https://github.com/facebookincubator/create-react-app/graphs/contributors?

I would also be happy to give it a sanity check.

examples/react/src/App.js
```diff
@@ -0,0 +1,148 @@
...
+
+class App extends Component {
+  state = {
+      nowShowing: ALL_TODOS,
+      editing: null,
+      newTodo: ''
+  }
+  componentDidMount () {
+    var setState = this.setState;
+    var router = Router({
```
gaearon: Can we use React Router? Ultimately it's what people are most likely to encounter in React ecosystem.

examples/react/src/App.js
```diff
@@ -0,0 +1,148 @@
...
+
+import TodoFooter from './Footer';
+import TodoItem from './TodoItem';
+import {ALL_TODOS, ACTIVE_TODOS, COMPLETED_TODOS} from './utils'
+
+const ENTER_KEY = 13;
+
+class App extends Component {
+  state = {
+      nowShowing: ALL_TODOS,
```
gaearon: Please format the project using [Prettier](https://prettier.github.io/prettier/).
In particular, we can see some mismatching indentation here.

examples/react/src/App.js
```diff
@@ -0,0 +1,148 @@
...
+    this.setState({newTodo: event.target.value});
+  }
+  handleNewTodoKeyDown (event) {
+    if (event.keyCode !== ENTER_KEY) {
+      return;
+    }
+
+    event.preventDefault();
+
+    var val = this.state.newTodo.trim();
```
gaearon: Please use `const` whenever possible, and fall back to `let`. Please don’t use `var`.

gaearon:
I think we can make this work, if the style issues are fixed. Right now the indentation is very inconsistent, but adopting Prettier will fix it.

@jchris Would you be interested in addressing the comments?

sindresorhus:
> From my perspective, a lot of beginners are going to use CRA rather than what we have in place here at present so I'm fine with moving ahead with that side of these changes.

👍

jchris:
Thanks for the review. This patch was opportunistic because I had it on my workbench for another project. I don't see it coming around to my front-burner again anytime soon. Maybe in a few months... if anyone is up for making these small changes that'd be lovely.

PR #1770, opened 2017-03-23
polymer2 initial checkin

jhaker:
This adds a polymer 2.0 example. Polymer 2.0 introduces a number of breaking changes which have been factored out of the original polymer example. 

The next round of updates will reduce polymer.dom api dependencies, upgrade class-based elements and replace customStyle instance properties.


addyosmani:
cc @wmginsberg - do you think it would be possible to get someone from the Polymer team reviewing this PR?

jhaker:
Looking for a todo ninja to help finish out; to help get approved. "Only those with write access to this repository can merge pull requests."

jorgecasar:
This isn't Polymer 2.x. It loads Polymer 2.x but components syntax is Polymer 1.x. You should use ES6 clases to define components and import polymer-element.html instead of polymer.html.

Issue #1566, opened 2015-12-28
New AngularJS HTML Only Implementation


dmitriz:
Please see here for more details:
https://github.com/dmitriz/todomvc/tree/ang-html/examples/angularjs-html

Before submitting a PR, I'd like to hear opinions and critics, and whether this would be appropriate here.

# Summary

## What?
- Perhaps one of the unusual AngularJS examples out there.
- Literally no JavaScript, apart from 2 tiny helper directives. 
- No controllers!
- Everything is inside HTML done with [Angular Expressions](https://docs.angularjs.org/guide/expression).

## Why?
- I wanted a proof of concept demonstrating the power of Angular Expressions. With some of the recent critics of Angular's apparent shortcomings, Angular's great and powerful parts become sidelined. HTML is the language of the Web, and AngularJS extends its capabilities.
- HTML files with maximum preserved structure may look friendlier to a designer or maintainer who just wants to make few text/style changes without getting deep into the code.
- The code can be smaller and faster to prototype.
- HTML can be a more natural place to put some logic, side-by-side with HTML-native accessibility and fallback support.

> It is very tempting to do everything in JavaScript, but it is not needed.
> When everything else fails, HTML is the thing users get - let's not deprive them of that.

> _[Chris Heilmann, The principles of a vanilla web diet](http://vanillawebdiet.com/)_

addyosmani:
Thanks so much for this contribution. I'm going through old issues and closing them out as we last updated the project in a major way in 2024 and want to ensure we're focused on only those issues still relevant. Thanks once again, feel free to reopen if still relevant

PR #1783, opened 2017-05-09
Angular pure html, no javascript!

dmitriz:
Please see here for more details:
https://github.com/dmitriz/todomvc/tree/ang-html/examples/angularjs-html

Before submitting a PR, I'd like to hear opinions and critics, and whether this would be appropriate here.
# Summary
## What?
- Perhaps one of the unusual AngularJS examples out there.
- Literally no JavaScript, apart from 2 tiny helper directives. 
- No controllers!
- Everything is inside HTML done with [Angular Expressions](https://docs.angularjs.org/guide/expression).
## Why?
- I wanted a proof of concept demonstrating the power of Angular Expressions. With some of the recent critics of Angular's apparent shortcomings, Angular's great and powerful parts become sidelined. HTML is the language of the Web, and AngularJS extends its capabilities.
- HTML files with maximum preserved structure may look friendlier to a designer or maintainer who just wants to make few text/style changes without getting deep into the code.
- The code can be smaller and faster to prototype.
- HTML can be a more natural place to put some logic, side-by-side with HTML-native accessibility and fallback support.

> It is very tempting to do everything in JavaScript, but it is not needed.
> When everything else fails, HTML is the thing users get - let's not deprive them of that.
> 
> _[Chris Heilmann, The principles of a vanilla web diet](http://vanillawebdiet.com/)_


dmitriz:
Also see the post https://github.com/tastejs/todomvc/issues/1566

FadySamirSadek:
@dmitriz Can you please rebase your fork and make your PR into one commit .Also I am not sure if this qualifies to the TodoMVC criteria as this is not the usual way people are going to use to write an angular app but i think it will make a very interesting tutorial.

PR #1784, opened 2017-05-24
Update ReadMe section on Editors for TypeScript-Angular example

CodeCharm:
Existing content was dated; brought to current


DanielRosenwasser:
:+1:

PR #1789, opened 2017-06-12
escape key pressed on blank todo deleted it

asutosh97:
escape key pressed on blank todo deleted it, rather than ignoring it just it does in all other versions.

### DIFF
--- a/examples/jquery/js/app.js
+++ b/examples/jquery/js/app.js
@@ -173,15 +173,14 @@ jQuery(function ($) {
 			var $el = $(el);
 			var val = $el.val().trim();
 
-			if (!val) {
-				this.destroy(e);
-				return;
-			}
-
 			if ($el.data('abort')) {
 				$el.data('abort', false);
 			} else {
-				this.todos[this.getIndexFromEl(el)].title = val;
+				if (!val) {
+					this.destroy(e);
+				} else {
+					this.todos[this.getIndexFromEl(el)].title = val;
+				}
 			}
 
 			this.render();


asutosh97:
please have a look at the travis error, I kept trying but can't figure out why the travis error is occuring.

examples/jquery/js/app.js
```diff
@@ -173,15 +173,14 @@ jQuery(function ($) {
 			var $el = $(el);
 			var val = $el.val().trim();
 
-			if (!val) {
-				this.destroy(e);
-				return;
-			}
-
 			if ($el.data('abort')) {
 				$el.data('abort', false);
```
FadySamirSadek: add back ` return;` as it avoids an extra call to` this.render()`; which is not needed

FadySamirSadek:
@ugitch @asutosh97  I agree that the return statement should be added back but I do not see anything wrong with the if statement can you please elaborate more ?

ugitch:
@FadySamirSadek @asutosh97 Nothing wrong, I just didn't see a need for nested clauses. Like this:
```javascript
if ($el.data('abort')) {
        $el.data('abort', false);
} else if (!val) {
        this.destroy(e);
        return;
} else {
        this.todos[this.indexFromElem(el)].title = val;
}
```

FadySamirSadek:
@ugitch Ah , you are absolutely right
@asutosh97 mind addressing this ?

luke0b1000:
How about this? I thought it kept with the flow. 
```
var $data = $el.data('abort');
if (!$data && !val) {
    this.destroy(e);
    return;
}
if ($data) {
    $el.data('abort', false);
}
```

ugitch:
@FadySamirSadek Can I fix this?

FadySamirSadek:
@ugitch Absolutely ,that will be great 💪👍

PR #1722, opened 2016-11-25
Update to Ember 2.10 and use HashLocation

cibernox:
This will (I think) fix the broken link to the EmberJS version.

It also updates to the version of the framework to 2.10

PR #1790, opened 2017-06-15
Bump to Ember 2.13.3

mixonic:
Mostly I've taken my lead from https://github.com/tastejs/todomvc/pull/1722

/cc @cibernox


examples/emberjs/todomvc/README.md
```diff
@@ -7,9 +7,12 @@
...
 ## Note for people updating this app.
 
-The `index.html` and the `assets` folder in the parent folder as simlinks into the items with the
-same names inside `dist`. The `dist` folder has to be checked in git and built for production.
+The `index.html` and the `assets` folder in the parent folder as simlinks into
+the items with the same names inside `dist`. The `dist` folder has to be checked
+in git and built for production.
 
-To work on this comment `<base href="/examples/ember-cli/index.html" />` in the `app/index.html`
-and uncommented it back before doing the production build.
```
mixonic: I've removed this documentation, as I believe it is out of date.

examples/emberjs/todomvc/config/targets.js
```diff
@@ -0,0 +1,10 @@
+/* eslint-env node */
+
+module.exports = {
+  browsers: [
+    'ie 11',
```
mixonic: Ember can be built back to `'ie 9'` here. I chose IE11 somewhat arbitrarily.

examples/emberjs/todomvc/config/environment.js
```diff
@@ -1,10 +1,10 @@
-/* jshint node: true */
+/* eslint-env node */
 
 module.exports = function (environment) {
 	var ENV = {
 		modulePrefix: 'todomvc',
 		environment: environment,
-		baseURL: null,
```
mixonic: Hm, did I miscalculate and this must be `null` to make the `index.html` file have no leading `/`?

mixonic: `index.html` paths look correct with `rootURL`. I believe `baseURL` is deprecated 🤔

examples/emberjs/todomvc/.travis.yml
```diff
@@ -1,25 +1,21 @@
 ---
```
FadySamirSadek: Why do we need a .travis.yml file inside this directory ?

examples/emberjs/todomvc/README.md
FadySamirSadek: I think this README.md file should be moved to the root directory of the example

examples/emberjs/todomvc/config/environment.js
FadySamirSadek: Yes, I think that baseURL is deprecated. The important thing is that it works 🙂

examples/emberjs/todomvc/package.json
```diff
@@ -16,28 +16,28 @@
...
+    "ember-export-application-global": "^2.0.0",
+    "ember-inflector": "2.0.1",
+    "ember-load-initializers": "^1.0.0",
+    "ember-resolver": "^4.0.0",
+    "ember-source": "~2.13.3",
+    "loader.js": "^4.2.3"
   },
   "engines": {
-    "node": ">= 0.12.0"
+    "node": ">= 6"
```
FadySamirSadek: I think we should lock this version WDYT?

zoltan-nz:
How about this PR? https://github.com/tastejs/todomvc/pull/1912

examples/emberjs/todomvc/README.md
zoltan-nz: Ah, I had the same feeling... I update the readme as Mixonic suggested and I move it to the root directory.

mixonic:
Closing in favor of https://github.com/tastejs/todomvc/pull/1912

Issue #1812, opened 2017-08-26
new app proposal: TODOMVC with Satori real-time platform


leebailey88:
I work for [Satori](https://www.satori.com/), and I feel that our framework would be a nice fit for the real-time section of your website.  Satori is a fast and free publish/subscribe system through which you can send and receive any JSON-serializable messages to/from any clients capable of using web sockets.  We also provide real-time message filtering and aggregation with an SQL-like language, and we also give you the ability to write your own streambot if you are want to run logic on the server when certain message conditions transpire.

If you are building a real-time app, Satori can be your backend similar to Firebase.  We have written some sample applications such as [chat](https://www.satori.com/apps/chat) and [paint](https://www.satori.com/apps/paint), but we'd very much love to write a TODO app to be hosted on your  site to help spread the word about how easy it is to get started with Satori.

FadySamirSadek:
@leebailey88 thank you for your contribution I will be reviewing your PR shortly

addyosmani:
Thanks so much for this contribution. I'm going through old issues and closing them out as we last updated the project in a major way in 2024 and want to ensure we're focused on only those issues still relevant. Thanks once again, feel free to reopen if still relevant

PR #1815, opened 2017-08-29
Implementing react-satori example

leebailey88:
I proposed this app in #1812 

This app builds on top of @petehunt 's react example app but uses Satori live data platform as the backend to exhibit real-time functionality.  With this app, you can open two separate browsers and the todos will be synced across each in real-time as changes are made in either browser.


leebailey88:
@addyosmani @passy would either of you be able to review my app or is there someone else I could ask to review?  Thanks very much for the consideration

PR #1816, opened 2017-08-30
jshint error: missing semicolon

serv-inc:
(minor)

https://github.com/tastejs/todomvc/blob/master/codestyle.md says that examples should be "JSHint valid"

### DIFF
--- a/examples/backbone_require/node_modules/todomvc-common/base.js
+++ b/examples/backbone_require/node_modules/todomvc-common/base.js
@@ -21,7 +21,7 @@
 				}
 			}
 			return object;
-		}
+		};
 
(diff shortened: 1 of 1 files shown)


passy:
Thanks, @serv-inc, but this is in a `node_module`, an installed dependency and not a part of the actual code of the application. 

The module exists here if you want to fix it: https://github.com/tastejs/todomvc-common

serv-inc:
Thanks so much, @passy, I did not know where to find it. The jshint error had been fixed in todomvc-common already.

serv-inc:
There is now a pull request for the missing semicolon: https://github.com/tastejs/todomvc-common/pull/17

PR #1822, opened 2017-09-13
Fix chromedriver lookup

kt3k:
I think we need this line to make selenium-webdriver recognize the chromedriver's path automatically.

```
chrome.setDefaultService(new chrome.ServiceBuilder(chromedriver.path).build());
```

And I pinned the chromedriver's versioin to 2.28.0 because of this issue ( https://github.com/samccone/drool/issues/32 ).


FadySamirSadek:
@kt3k I believe this is no longer needed as we completely changed our test suite to cypress but your help in completing the new test suite will be so much appreciated , thanks for this PR

Issue #469, opened 2013-02-25
Backbone implementation: Rendering executed twice 


jdittrich:
In the implementation using Backbone, the rendering using /views/todo.js,  render: function () {...} is triggered twice if a new item is created: First on the actual creation, than triggered by a change event. This is seemingly caused by the implementation of backbone-localstorage, which adds an attribute to the newly created model (see http://stackoverflow.com/questions/10228267/backbone-js-fires-render-twice-on-collection-add) causing the change to fire right after the item is added to the collection. Would it be possible to avoid that behaviour or to add a comment pointing out to it? 
I suppose many people use the example as a in-depth intro via debugging and at least I assumed that the double rendering was needed for some reason – which I could not find for hours...

PR #1826, opened 2017-09-18
Backbone: Listen for changes to individual attributes

shiyu-zhang:
Backbone implementation used to trigger render function twice if a new item is created as reported in #469. A workaround is to filter out the render invocation caused by 'id' change as #755 did. 

Maybe it's more decent to fix this issure by binding listeners to specific attributes change of the model. In this way, we only listen to the model attributes that we really care and ignore the 'id' change. It's also  more efficient to write specific listeners for different model attributes ('completed' and 'title') change instead of re-rendering the whole todo item once the mode changes, I think. The performance of backbone suite in Speedometer-2.0 improved 8% after applying this patch.

PS: You can try https://github.com/TianyouLi/javascript-benchmarks/tree/master/Speedometer-2.0, deploy it to your web server, launch the index.html, click "frameworks", to check the framework you would like to run.


pdeng6:
There is a discussion about keyed/nonkeyed problem.
https://bugs.webkit.org/show_bug.cgi?id=176586

PR #1830, opened 2017-10-21
Focus caret at end of input when in editMode

starogre:
I updated the line of code that was supposed to focus the caret at the end of the input line when in editMode. It no longer does what I assume it was supposed to do, so I created a way for it to work. This makes it easier to backspace or add more to your Todo. It will also make the code easier for beginners to understand line by line.


FadySamirSadek:
@ugitch can you please create an issue and a PR to solve it if you are intrested

ugitch:
Sure thing, @FadySamirSadek!

PR #1837, opened 2017-11-15
Add a Git .mailmap with my new name

mgol:
In this way my past contribution is mapped correctly.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,1 @@
+Michał Gołębiowski-Owczarek <m.goleb@gmail.com>


mgol:
`master` is failing, test failures are unrelated to this PR.

samccone:
Is this critical for you? Since github does not support this https://github.com/holman/ama/issues/551, it is only useful for local git clients so I do not see a major win landing this.

Thanks

mgol:
Local Git clients also include IDEs etc. so I find this useful. But it's definitely not critical; I just prefer this data to be correct but if you'd rather not add a new file, I'm fine with not landing this PR.

samccone:
I am find landing it after reflecting more.

Thanks so much for the PR!

PR #1846, opened 2018-01-29
w3view - minimalistic view layer lib

vitalydmitriev1970:
Hello, i think it is new way, that looks like very old. Thin layer on top of the DOM API can solve probably all problems :)


vitalydmitriev1970:
js lint was failed on recomended  director.js :)

PR #1854, opened 2018-02-28
Typescript-Vue

victorgarciaesgi:


victorgarciaesgi:
Yes still want to! I just need to pull master right?

Fruitseller:
@victorgarciaesgi I think [rebasing](https://git-scm.com/book/en/v1/Git-Branching-Rebasing) would be a cleaner aproach for getting your PR up to date.
You should also look if you codebase fulfills the app [requirements](https://github.com/tastejs/todomvc/blob/master/app-spec.md).

And I think it would be great if you could write something about your motivations to add vue in TS to todomvc and also what the benefits are when its merged.😊

PR #1867, opened 2018-03-18
Update cypress

samccone:
Also fix react and assorted tests that place elements on top of the check box toggle (which can cause the tests to fail since they think there is an element on top)

--

Also migrate the cypress project to be under tastejs instead of cypress, and make sure to record videos and upload results to cypress

https://dashboard.cypress.io/#/projects/n4ynap/runs

\o/

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -1,7 +1,6 @@
 {
-  "videoRecording": false,
   "baseUrl": "http://localhost:8000/examples/",
-  "projectId": "569hk8",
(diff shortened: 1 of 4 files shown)


FadySamirSadek:
We can place the cypress badge beside travis on the readme.md file to link to the cypress dashboard 
[Cypress.io tests](https://dashboard.cypress.io/#/projects/n4ynap/runs)

PR #1870, opened 2018-03-28
Enable npm ci , package-lock.json

FadySamirSadek:
npm ci is a new install command works as npm install but is much faster and made specially for CI systems so it will be more reliable and faster which will decrease the time needed to verify and test a new PR which in return will help new contributors.

Resources:
http://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,10 @@ env:
     - GH_PROJECT_NAME: todomvc
     - CHROME_PATH: '/home/travis/build/tastejs/todomvc/chrome-linux/chrome'
 
+#update npm to latest stable version
+before_install:
+  - npm i -g npm
+
 # common build steps for each test job
 defaults: &defaults
   script:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -31,6 +31,10 @@ env:
     - GH_PROJECT_NAME: todomvc
     - CHROME_PATH: '/home/travis/build/tastejs/todomvc/chrome-linux/chrome'
 
+#update npm to latest stable version
+before_install:
+  - npm i -g npm
```
samccone: this seems kinda weird :)

samccone:
LGTM minus 1 comment

.travis.yml
FadySamirSadek: I added this because the default npm version TravisCI is using is npm v5.1 while npm CI needs npm v5.8.
This is the only way I have in mind to change the default npm version to 5.8 as per https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Travis-CI-uses-npm
WDYT?

samccone: hmm alright mind setting it to something exact then npm i -g npm@X.X.X ?

FadySamirSadek: @samccone done

samccone:
Thanks @FadySamirSadek

Issue #1801, opened 2017-07-25
jQuery blank-name deletion on escape-key bug


asutosh97:
The Bug:- In jQuery version of the app, if we rename any todo to blank-name (""), and then press escape, then instead of ignoring the changes and reverting back to the oldname as it does in other versions of the app, the todo gets deleted in the jQuery version.

I am interested in fixing the fixing the issue, but I'm new to open source contribution. So, can you please guide me how to proceed.

hejob:
Spotted the same bug.
Seemingly App.update() is also called after escape key press, and not performing right decisions in the function caused the problem.
Moving the if statement for ($el.data("abort")) before the if statement of (!val) and this.destroy(e) command fixed the bug for me.

FadySamirSadek:
@asutosh97 Are you still interested in picking this one up ?
check the [contribution guide](https://github.com/tastejs/todomvc/blob/master/contributing.md) and if you need extra help , I will be happy to chat and work with you to ship this one .
Nice catch , keep up the good work 👍

FadySamirSadek:
@addyosmani @samccone  this one should be marked as bug and as good for first time contributors.

FadySamirSadek:
There is an open PR already at #1789

FadySamirSadek:
closed as per #1876

PR #1876, opened 2018-04-18
jQuery - todo with empty value deletes todo #1789 #1801

ugitch:
When you're updating todo, you shouldn't leave out the value. If doing so, update operation should be aborted.


ugitch:
You're welcome, @FadySamirSadek

PR #1906, opened 2018-06-29
Fix cypress test run for vanilla-es6

### LINKED ISSUE #1903: Fix cypress test run for vanilla-es6
Re #1866 

The closure compiler optimizes the item "completed" field to "c" which breaks cypress test checks. It could be configured to not do that. However I think a better solution here is to switch to webpack and babel as that it probably more widely used.

Another issue is that the item is not wrapped in a div.view.

And this framework uses blur to complete edit.

gj262:
wrap item in a 'view' div for inline edit & to pass cypress test
use webpack & babel instead of closure compiler to avoid "completed" shortened to "c"
update to latest todomvc-*
vanilla-es6 saves edit on blur

Fixes #1903


examples/vanilla-es6/package.json
```diff
@@ -1,14 +1,18 @@
 {
   "private": true,
   "scripts": {
-    "compile": "java -jar node_modules/google-closure-compiler/compiler.jar -O ADVANCED --language_in=ES6_STRICT --new_type_inf --js_output_file='dist/bundle.js' 'src/**.js' -W VERBOSE",
```
samccone: if you drop advanced no renaming will happen ... no need to switch to webpack :)

gj262: Yeah.. I was really hoping to avoid a java install :(. And I think it is safe to say most people are using webpack/babel and not closure. So a switch will make this easier for most to maintain. No?

FadySamirSadek:
closed for #1907

PR #1907, opened 2018-06-29
fix cypress test for vanilla-es6

### LINKED ISSUE #1903: Fix cypress test run for vanilla-es6

gj262:
Like #1906 but keeps the closure compiler. 

wrap item in a 'view' div for inline edit & to pass cypress test
remove advanced optimization from closure options to stop turning "completed" to "c"
update to latest todomvc-*
vanilla-es6 saves edit on blur

Fixes #1903


samccone:
Thank you for the PR!

Issue #1886, opened 2018-05-25
 Create a clear guideline to help us decide what makes a framework mature, production ready and fit to be added to TodoMVC


FadySamirSadek:
Resources : 
-https://addyosmani.com/blog/front-end-choice-paralysis/
-https://github.com/tastejs/todomvc/blob/master/contributing.md#considerations-before-submitting-a-new-app

Ahmed-90:
Can we say a framework should: (as a very basic example)

- Be at least 1 year old. (git commits?)
- Have at least 1k stars. (proof of public interest)
- Be compatible with modern standards and techniques like bundlers, code splitting, and offline first approaches like PWAs, services workers etc...
- Backward compatible for 2 older versions of at least 2 major browsers. (overkill?) 
- Should be unique/different and not just an enhanced/features rich fork of an existing framework.
      * If this line is a valid guideline what about things like nuxt/vue, next/react? 

---

Am I on the right track here?  
Should we mention licensing, opensource, etc...?

FadySamirSadek:
@Ahmed-90 this is very similar to what I had in mind. 

> Be at least 1 year old

Completely agree

> Have at least 1k stars. (proof of public interest)

Also agree but I think we should make it 5k instead of 1k almost all the widely used projects are way passed this number React now has 104K and vue has 105K any framework that has been around for a while is going to have more than 5k stars if it is widely adopted which is exactly what we need.

> Be compatible with modern standards and techniques like bundlers, code splitting, and offline first approaches like PWAs, services workers etc...

Also agree but we should make a comprehensive list as the goal of these guidelines is to not think when adding a new framework and make it as close to a yes-no question as possible either the framework checks all the points or not.

> Backward compatible for 2 older versions of at least 2 major browsers

Not sure about backward compatibility but should definitely support recent versions of all major browsers.

>Should be unique/different and not just an enhanced/features rich fork of an existing framework.

I think its okay to have multiple examples and multiple implementations for the same framework as long as the framework is widely used and these implementations are considerable options for people starting new projects then we should have them.

Regarding licensing I think the framework must have a version that is free and open source no matter what the license is.

WDYT?

Ahmed-90:
> Also agree but I think we should make it 5k instead of 1k

Initially, I actually thought about a high number, but I had the notion of "what about the shiny new projects", but yeah you're right ~~hype~~ stars are easy to generate, 5k seems reasonable.

[...]

PR #1910, opened 2018-07-13
New examples guidelines

FadySamirSadek:
solves #1886


contributing.md
```diff
@@ -22,6 +22,13 @@ Please bear in mind that your project should have reached a certain level of
 maturity before we can include it. If you haven't settled on a relatively stable
 API, you should probably wait a bit longer before submitting your example.
 
+### A good app that we are looking for should check most if not all of the items below :
+
+- [x] The framework/stack you are using must have a version that is free and open source
+- [x] The framework/stack you are using is at least 1 year old.
+- [x] The framework you are using should have at least 5k stars if it is hosted on Github or equivalent on other platforms.
+- [x] Your app should be able to work on recent versions of all major browsers.
+- [x] Your app should bring something truly new to the table, but minor variations of existing frameworks , languages or existing TodoMVC apps.
```
Ahmed-90: I believe that's a typo? "~~but~~ **not** minor variations..."?

FadySamirSadek: Will fix it, thanks 👍

FadySamirSadek:
@whitefire0 I do not think an exhaustive list of features is the solution I think the first four guidelines are yes-no questions which is what we need but the last one just means your example has to be different because people can always use the same stack in different ways. I think we will always have to discuss and think if an app is a good addition or not but the first 4 guidelines  are going to help us a lot in the decision making process

FadySamirSadek:
@Ahmed-90 Fixed the typo and rephrased the statement.
 can you please have a look? Thanks 🙂

Ahmed-90:
@FadySamirSadek Looks good to me. 👍 
BTW I still think that this list should be enhanced with a technical (lighthouse like) checklist to ensure standardization.

FadySamirSadek:
@Ahmed-90 I think this should be added incrementally. I think we should add lighthouse to our CI system but this can be added in an hour in the future. Will open an issue for it and feel free to submit a PR.

PR #1925, opened 2018-08-04
Io.js example

arodic:
Here is a todomvc implementation with my recently published library [Io.js](https://github.com/arodic/io).

I am aware this library does not meet criteria for inclusion (no stars, not working in all browsers etc.) but I believe it does include some novel ideas and I'm hoping to get early feedback as I'm moving on with development.


arodic:
I'm still working on

PR #1931, opened 2018-08-18
circularJS implementation

PitPik:
Introducing circularJS.
Please read README-FIRST.txt which explains all about this pull request.


FadySamirSadek:
@PitPik Thank you for your PR, we are currently shorthanded when it comes to maintaining examples also we have the 5000 stars rule for a reason which is to not confuse new and upcoming developers by the number of choices. Though I am open to adding a link to your example to our wiki till your framework meets the requirements needed to be included.

PitPik:
Hi @FadySamirSadek 
thanks for the answer. The link to my implementation would be: https://pitpik.github.io/circularjs/todo/ in case you consider the link in your wiki.
Cheers

PR #1932, opened 2018-08-23
remove frameworks that do not comply with the new TodoMVC guidelines

FadySamirSadek:


nelsonic:
@FadySamirSadek _totally_ agree with "pruning" the frameworks that do not comply with guidelines ... could you please share a link to the guidelines so beginners are aware which ones you are referring to in this PR? (thanks!)

Issue #1971, opened 2019-04-15
Counter of active todos should be bold


ugitch:
From the Application Specification, section Counter:
Make sure the number is wrapped by a `<strong>` tag. Also make sure to pluralize the `item` word correctly: `0 items`, `1 item`, `2 items`. Example: **2** items left

coolbeatz71:
Hey @ugitch I think this issue is already solved. Particularly on the react example.

PR #1972, opened 2019-04-15
Update index.css. #1971

ugitch:
Removing redundant styling.


flashdesignory:
CSS updates of the todomvc app should happen [here](https://github.com/tastejs/todomvc-app-css).

PR #1984, opened 2019-07-21
Examples on How to Convert Frameworks to ES6

infolock:
First, todo mvc is a god send.   Thank you all very much for this great library!

Secondly, this PR has some explaining to do if it is to be considered for merge.  Please also checkout the `README.md` found in this PR for even more information.

**What is this PR?**
This PR contains examples of how to convert a framework to ES6+using the patterns that they promote.  

**Where is the code?**
The code can be found in the `es6-examples` folder

**Why does only _some_ of the functionality work in this PR?**
This was actually done intentionally for the following reasons:
1. Allows developers to see how this type of thinking can be done (via creating simple solutions first - and _then_ creating a more solid architectural solution _after_)
2. Allows developers to attempt to get hands on experience actually attempting to solve the problems and write the code themselves - which encourages stronger learning techniques
3. Forces the developers to switch back to the framework, understand what it is doing, and then attempt to decypher and re-create it using ES6+!

If required to make a full example out of these two, just ask and I'll put it in place.   Shouldn't take but a couple hours anyways.

**What value does this PR provide?**
The values for this (imo) are endless.  Some ideas that this promotes:
- learning frameworks as a means to understanding solid patterns and naming conventions
- how to take a pattern and apply it with native ES6+
- understanding how to do this in the real world if they decide to go down this course
- educates engineers to rely more on the things that ES6+ now provides that ES5 and before didn't
- showcases how much cleaner native JS is today with ES6+

**Why is this in its own folder?**
Well - i didn't want to pollute the `examples` folder with the same folder names twice.   It just felt cleaner this way...

More information can be found in the `README.md` within this PR.  I can also add more clarity as needed.   If the examples need to be better organized, written, or completed - don't hesitate to request so.  My own personal opinion is that it would take away from the very thing this PR is attempting to do, which is educate engineers by forcing them to apply some hands-on experience to complete it (since the work to do so has already been done in the framework example itself - leaving only translating it to native JS)

**How does it work?**
Its a pretty simple, straightforward solution really..

1. NodeJS is used to load the environment.  This may go against the todomvc way - but i guess that's up  to  you all to decide
2. Uses webpack for backwards transpilation to es5
3. Has hot-reloading built-in and ready to go

**To run each example:**
  - `cd from-backbone && npm i && npm run start` and open your browser to [http://localhost:3000/]( http://localhost:3000/)
  - `cd from-react && npm i && npm run start`  and open your browser to [http://localhost:3000/]( http://localhost:3000/)

**Recent Change Log**
- added some missing contributors
- squashed commits
- revert accidental changes to pre-existing items (`package.json`, `package-lock.json`, `examples/react` and `examples/backbone`)
- build issues corrected


infolock:
I'm gonna close this PR and repoen it later.  After reading through it a little more, I think it should be a little cleaner and more straight forward with an actual complete solution first.   tbc

PR #2089, opened 2019-10-23
Feature/descriptive titles

andrehgdias:
Pretty simple changes: 

- The h1 title element sets its inner HTML with page title (in  examples/react/js/app.jsx)
- Css property font-size and top, from .todoapp h1, altered to fit its page title text ( examples/react/node_modules/todomvc-app-css/index.css).

Just need to know if it's okay to do these alterations.

### DIFF
--- a/examples/react/js/app.jsx
+++ b/examples/react/js/app.jsx
@@ -154,7 +154,7 @@ var app = app || {};
 			return (
 				<div>
 					<header className="header">
-						<h1>todos</h1>
+						<h1>{document.title}</h1>
 						<input
 							className="new-todo"
 							placeholder="What needs to be done?"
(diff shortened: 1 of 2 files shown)


lyndseypadget:
This looks fine to me, despite the failing build (@FadySamirSadek, can you advise on missing env variables?)

If a core team member agrees, I would prefer updating all example apps in this way in one PR as opposed to a separate PR for each.  You could continue updating this branch until they are all done AFAIC.

andrehgdias:
That's ok to me, but I can work only on a few of the examples since I'm still studying JavaScript programming and its huge ecosystem.

FadySamirSadek:
> This looks fine to me, despite the failing build (@FadySamirSadek, can you advise on missing env variables?

The test suite is currently broken but I'm working to fix that.

> If a core team member agrees, I would prefer updating all example apps in this way in one PR as opposed to a separate PR for each. You could continue updating this branch until they are all done AFAIC.

I agree this should be done in a single PR. If you can work with @andrehgdias on that, it would  be great.

PR #2104, opened 2020-02-17
Update avatar of member

hong4rc:
Use URL with GitHub API instead of `gravatar`

image


flashdesignory:
This has been resolved in a similar approach with the last update. 
Closing this pr for now.

PR #2120, opened 2020-07-14
updating mithril

boazblake:
this pr updates the mithril.js example to the latest version.


mocktarissa (changes requested):
You committed too many files and there are too many files to be able to accept your changes

PR #2144, opened 2021-05-16
Add new Solid.js example

BasixKOR:
* Please use squash merge.

Hi, I created TodoMVC with Solid.js and I think it's worth sharing with y'all, so I'm proposing this to as a new app example.

image

It passes all the tests as you can see above.

# Reason to propose the TodoMVC in Solid.js
* It is one of the most perfomant UI libraries out there, according to https://github.com/krausest/js-framework-benchmark.
* It is based on DOM Expressions instead of VDOMs.

# Notes
* Since JSHint doesn't seems to support latest JavaScript features such as JSX and modules, it was hard to match code styles using JSHint.
* I bulit up custom Rollup config to work nicely with current infra. Probably a lot of space for improvement.


examples/solid/src/components/TodoList.jsx
```diff
@@ -0,0 +1,31 @@
+import { Switch, useContext } from "solid-js";
```
Progyan1997, line 1: Import `For`?

examples/solid/src/components/App.jsx
```diff
@@ -0,0 +1,34 @@
+import { For, Show, useContext } from 'solid-js';
```
Progyan1997, line 1: `For` is not being used

examples/solid/src/components/TodoItem.jsx
```diff
@@ -0,0 +1,46 @@
+import { createEffect, createSignal, useContext } from "solid-js";
```
Progyan1997, line 1: `createEffect` not being used

PR #2152, opened 2021-07-05
ops_task_approach_two

bala89thangavel:
reverse modified into for loop.we can use lodash also. forEachRight


flashdesignory:
Closing since the React example has been updated: https://github.com/tastejs/todomvc/pull/2202

PR #2178, opened 2022-08-03
Add Example using Lit

christian-bromann:
[Lit](https://lit.dev/) is a great framework for building apps based on web components. This patch adds a TodoMVC example using this framework + Redux for data management. If desired I can remove the redux part.

### DIFF
--- a/examples/lit/js/components/TodoItem.js
+++ b/examples/lit/js/components/TodoItem.js
@@ -1,0 +1,97 @@
+import { connect } from 'pwa-helpers'
+import { html, css, LitElement, unsafeCSS } from 'lit';
+import { classMap } from 'lit/directives/class-map.js';
+
+import { store, setEdit, setTodo, toggleTodos, leaveEdit, deleteTodo } from '../store.js'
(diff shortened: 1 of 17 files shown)


examples/lit/js/components/TodoItem.js
```diff
@@ -0,0 +1,97 @@
...
+        }
+        `,
+        ...baseCSSRules.map((r) => unsafeCSS(r)),
+        ...appCSSRules.map((r) => unsafeCSS(r))
+    ]
+
+    get isEditing () {
+        const state = store.getState()
+        return this.todo.id === state.editingTodo
+    }
```
julienw, line 29: I think it would be better to `connect` the TodoList component, and pass `isEditing` as a boolean property. 
This way we wouldn't rerender all the items (see `requestUpdate` below) when the state of just one item changes, or even the state of any other thing that's not related to these items.

Note that I know very little about Lit so I may be off here.

flashdesignory:
closing in favor of https://github.com/tastejs/todomvc/pull/2196

PR #2193, opened 2023-09-11
pin node version in package.json 

flashdesignory:
This ensures that a node version is used that is compatible with the current node packages.
NVM will use node 8.17.0 and you'll be able to install and spin up the repo locally.

```
npm install
npm run gulp serve
```

After that step, open browser and navigate to: http://localhost:8080/

@kara

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "todomvc",
-  "version": "0.1.1",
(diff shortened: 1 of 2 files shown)


kara (approved):
LGTM

addyosmani:
👍🏻

PR #2194, opened 2023-09-13
Add Svelte to Examples

flashdesignory:
This pr adds a Svelte example.

@kara


examples/svelte/src/App.svelte
```diff
@@ -0,0 +1,66 @@
+<script>
+    import { router } from './router.js';
+    import { uuid } from './utils.js';
```
Rich-Harris, line 3: we can use the platform here

```suggestion
```

examples/svelte/src/App.svelte
```diff
@@ -0,0 +1,66 @@
...
+
+    import Header from './Header.svelte';
+    import Footer from './Footer.svelte';
+    import Item from './Item.svelte';
+
+    import "./app.css";
+    import "todomvc-app-css/index.css";
+    import "todomvc-common/base.css";
```
Rich-Harris, line 11: I had previously left these out so they didn't affect the bundle size of the app itself (since some people are concerned with that), but if this is commonly bundled with the other implementations then 👍

examples/svelte/src/App.svelte
```diff
@@ -0,0 +1,66 @@
...
+
+    let currentFilter = "all";
+    let items = [];
+    let editing = null;
+
+    function addItem(event) {
+        items.push({
+            id: uuid(),
```
Rich-Harris, line 19:
```suggestion
            id: crypto.randomUUID(),
```

examples/svelte/src/App.svelte
```diff
@@ -0,0 +1,66 @@
...
+        items.splice(index, 1);
+        items = items;
+    }
+
+    function toggleAllItems(event) {
+        const checked = event.target.checked;
+        items = items.map((item) => ({
+        ...item,
+        completed: checked,
+        }));
```
Rich-Harris, line 36:
```suggestion
        items = items.map((item) => ({
            ...item,
            completed: checked,
        }));
```

examples/svelte/src/App.svelte
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    function removeCompletedItems() {
+        items = items.filter((item) => !item.completed);
+    }
+    
+    router(route => currentFilter = route).init();
```
Rich-Harris, line 43: This should really be happening inside an `onMount` callback. Doesn't really matter for this demo since it only runs client-side, but in general we avoid doing anything browser-specific during component initialisation

```suggestion
    onMount(() => {
      router(route => currentFilter = route).init();
    });
```

(needs `onMount` to be imported from `svelte` obvs)

examples/svelte/src/App.svelte
```diff
@@ -0,0 +1,66 @@
...
+            <input id="toggle-all" class="toggle-all" type="checkbox" on:change={toggleAllItems} checked={numCompleted === items.length} />
+            <label for="toggle-all">Mark all as complete</label>
+        </div>
+        <ul class="todo-list">
+            {#each filtered as item, index (item.id)}
+                <Item bind:item editing={editing} on:removeItem={() => removeItem(index)} />
+            {/each}
+        </ul>
+
+        <Footer numActive={numActive} currentFilter={currentFilter} numCompleted={numCompleted} on:removeCompletedItems={removeCompletedItems} />
```
Rich-Harris, line 64:
```suggestion
        <Footer {numActive} {currentFilter} {numCompleted} on:removeCompletedItems={removeCompletedItems} />
```

examples/svelte/src/App.svelte
```diff
@@ -0,0 +1,66 @@
...
+
+{#if items.length > 0}
+    <main class="main">
+        <div class="toggle-all-container">
+            <input id="toggle-all" class="toggle-all" type="checkbox" on:change={toggleAllItems} checked={numCompleted === items.length} />
+            <label for="toggle-all">Mark all as complete</label>
+        </div>
+        <ul class="todo-list">
+            {#each filtered as item, index (item.id)}
+                <Item bind:item editing={editing} on:removeItem={() => removeItem(index)} />
```
Rich-Harris, line 60: In the original implementation `editing` is the index of the todo that's currently being edited. Here, it's a per-component piece of state, and it looks like we're relying on the `blur` event to ensure that it's only true for one item at a time. I think we can therefore remove it

```suggestion
                <Item bind:item on:removeItem={() => removeItem(index)} />
```

examples/svelte/src/App.svelte
```diff
@@ -0,0 +1,66 @@
...
+    import Footer from './Footer.svelte';
+    import Item from './Item.svelte';
+
+    import "./app.css";
+    import "todomvc-app-css/index.css";
+    import "todomvc-common/base.css";
+
+    let currentFilter = "all";
+    let items = [];
+    let editing = null;
```
Rich-Harris, line 15: per comment elsewhere, this isn't needed

```suggestion
```

examples/svelte/src/Item.svelte
```diff
@@ -0,0 +1,55 @@
+<script>
+    import { createEventDispatcher, tick } from 'svelte';
+
+    export let item;
+
+    export let editing = false;
+    const dispatch = createEventDispatcher();
```
Rich-Harris, line 7: `editing` shouldn't be a prop, per other comments, and convention dictates (or at least suggests) putting things like event dispatchers after props but before component state

```suggestion
    const dispatch = createEventDispatcher();

    let editing = false;
```

examples/svelte/src/Item.svelte
```diff
@@ -0,0 +1,55 @@
...
+        editing = false;
+    }
+
+    async function focusInput(element) {
+        await tick();
+        element.focus();
+    }
+</script>
+
+<li class="{item.completed ? ' completed' : ''}{editing ? ' editing' : ''}">
```
Rich-Harris, line 41:
```suggestion
<li class:completed={item.completed} class:editing>
```

examples/svelte/src/utils.js
```diff
@@ -0,0 +1,7 @@
+export function uuid() {
+    return "xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx".replace(/[xy]/g, function (c) {
+        var r = (Math.random() * 16) | 0,
+            v = c === "x" ? r : (r & 0x3) | 0x8;
+        return v.toString(16);
+    });
+}
```
Rich-Harris, line 7: per other comments, this module is unnecessary, we can just use `crypto.randomUUID()` nowadays

```suggestion
```

examples/svelte/src/App.svelte
flashdesignory: We used this function before, but opted out - [here](https://github.com/WebKit/Speedometer/issues/136#issuecomment-1553880792)'s the issue from Speedometer

examples/svelte/src/App.svelte
flashdesignory: Yeah - other apps typically import the base.css as well.

examples/svelte/src/App.svelte
flashdesignory: ah - now it makes sense, after reading all comments - please see my note below

examples/svelte/src/App.svelte
flashdesignory: this is also used to assign a [conditional class](https://github.com/tastejs/todomvc/blob/5d8525402426aedce9448c2fa7496b224e4c0a36/examples/svelte/src/Item.svelte#L41) for styling and a [conditional render](https://github.com/tastejs/todomvc/blob/5d8525402426aedce9448c2fa7496b224e4c0a36/examples/svelte/src/Item.svelte#L49).

examples/svelte/src/Item.svelte
flashdesignory: moved dispatch to the top!

examples/svelte/src/App.svelte
Rich-Harris: Oh, interesting. What about using `crypto.randomUUID()` for TodoMVC itself and using the `uuid` helper for Speedometer? Or do the implementations need to be identical?

examples/svelte/src/App.svelte
Rich-Harris: Yes, but it's state that's internal to `Item.svelte`. The prop here isn't doing anything (it never changes in `App.svelte`) and can be removed

Rich-Harris (approved):

examples/svelte/vite.config.js
```diff
@@ -0,0 +1,7 @@
+import { defineConfig } from 'vite';
+import { svelte } from '@sveltejs/vite-plugin-svelte';
+
+export default defineConfig({
+	plugins: [svelte()],
+    base: "examples/svelte/dist/"
```
dummdidumm, line 6:
```suggestion
	base: "examples/svelte/dist/"
```

examples/svelte/src/Footer.svelte
```diff
@@ -0,0 +1,30 @@
...
+<footer class="footer">
+    <span class="todo-count">
+        <strong>{numActive}</strong>
+        {numActive === 1 ? "item" : "items"} left
+    </span>
+
+    <ul class="filters">
+        <li><a class={currentFilter === "all" ? "selected" : ""} href="#/">All</a></li>
+        <li><a class={currentFilter === "active" ? "selected" : ""} href="#/active">Active</a></li>
+        <li><a class={currentFilter === "completed" ? "selected" : ""} href="#/completed">Completed</a></li>
```
dummdidumm, line 24:
```suggestion
        <li><a class:selected={currentFilter === "all"} href="#/">All</a></li>
        <li><a class:selected={currentFilter === "active"} href="#/active">Active</a></li>
        <li><a class:selected={currentFilter === "completed"} href="#/completed">Completed</a></li>
```

dummdidumm (approved):

FadySamirSadek (approved):

PR #2198, opened 2023-09-19
Add React Redux to Examples

flashdesignory:
This version has been lifted from Speedometer and follows the WebKit styles guide.

@kara


examples/react-redux/package.json
```diff
@@ -0,0 +1,48 @@
...
+        "webpack-cli": "^5.0.1",
+        "webpack-dev-server": "^4.11.1",
+        "webpack-merge": "^5.8.0"
+    },
+    "dependencies": {
+        "classnames": "^2.2.5",
+        "prop-types": "^15.8.1",
+        "react": "^16.14.0",
+        "react-dom": "^16.14.0",
+        "react-redux": "^7.2.6",
```
FadySamirSadek, line 42: Same for react-redux, the latest version is 8.1.2 as of the time of this review

examples/react-redux/src/components/footer.jsx
```diff
@@ -0,0 +1,45 @@
+import { Component } from "react";
+import classnames from "classnames";
+import PropTypes from "prop-types";
+
+import { SHOW_ALL, SHOW_COMPLETED, SHOW_ACTIVE } from "../constants/todo-filters";
+
+export default class Footer extends Component {
```
FadySamirSadek, line 7: Since the more recent function components are more used and are recommended over class components by the react core team shouldn't we be using them instead of class components? I am aware that this is a big change so it might make more sense to just merge this now and update it later on. 
Would love to know your opinion regarding this

examples/react-redux/webpack.common.js
```diff
@@ -0,0 +1,43 @@
+const path = require("path");
```
FadySamirSadek, line 1: I don't have a strong opinion regarding this but since this is a very simple app shouldn't we use something like parcel to avoid config files and make the example more simpler

examples/react-redux/public/index.html
```diff
@@ -0,0 +1,19 @@
...
+        <meta name="description" content="A TodoMVC written in React-Redux." />
+        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+        <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+        <title>TodoMVC: React-Redux</title>
+    </head>
+    <body>
+        <section class="todoapp" id="root"></section>
+        <footer class="info">
+            <p>Click on input field to write your todo.</p>
+            <p>At least two characters are needed to be a valid entry.</p>
```
FadySamirSadek, line 14: These instructions are not added to other TodoMVC implementations, I don't mind adding them but I would like us to be consistent

FadySamirSadek (commented):
Thank you so much @flashdesignory for the amazing work, I'm super excited to merge it alongside other Speedometer examples. I've left a few comments on certain aspects and am curious about your perspective on them. While I believe some of my suggestions could enhance our work, I recognize that they might not align with our immediate objective of updating the examples. Please view my comments as open-ended suggestions for discussion rather than definitive change requests.

examples/react-redux/package.json
```diff
@@ -0,0 +1,48 @@
...
+        "style-loader": "^3.3.1",
+        "webpack": "^5.75.0",
+        "webpack-cli": "^5.0.1",
+        "webpack-dev-server": "^4.11.1",
+        "webpack-merge": "^5.8.0"
+    },
+    "dependencies": {
+        "classnames": "^2.2.5",
+        "prop-types": "^15.8.1",
+        "react": "^16.14.0",
```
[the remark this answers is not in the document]
flashdesignory: When we updated the Speedometer specific versions, we looked at httpArchive and NPM stats to determine, which versions are most widely used on the web today. These queries were done around March '23, so they are slightly outdated now.

examples/react-redux/package.json
flashdesignory: Same as comment above, 7.2.6 was the most popular version at the time we chose what version to upgrade to.

examples/react-redux/src/components/footer.jsx
flashdesignory: For this build, we chose a traditional class pattern to implement the todoMVC application. It's still used widely and I believe it still represents a pattern that is useful to the community. 

There will be a React specific version that will use functional components and hooks to represent a more modern approach.

examples/react-redux/webpack.common.js
flashdesignory: We could, this was just what we used for Speedometer. The bundler was less of a focus point for that project. 
Since it works, I think this is not a show stopper, maybe a follow up task to update?

examples/react-redux/public/index.html
flashdesignory: I simplified for now - please let me know if you'd like me to change it 😄

flashdesignory:
> While I believe some of my suggestions could enhance our work, I recognize that they might not align with our immediate objective of updating the examples.

Hi @FadySamirSadek - thank you for your review 🙏 

For Speedometer, we made some decisions in regards to what version to upgrade to and what type of pattern to use to implement the todoMVC application. The big callout here is that we created two React versions. 

1. React Redux, which uses the most commonly used React and React-Redux version in combination with class components. 
2. React only, with functional components and hooks with a more recent React version (pr to come)

There could be an argument made to use React Toolkit vs React Redux, but that could be an additional community contribution that might be interesting to see. 

There are multiple ways a React + Redux todoMVC app could be created. This is just one version that we chose that worked for our purpose.

FadySamirSadek (approved):
LGTM

If @addyosmani and @kara green light this we can go ahead and merge it

Issue #2199, opened 2023-09-19
Upstream Speedometer 3.0 TodoMVC apps to TodoMVC


addyosmani:
Over the last year, Chrome and other browser vendors have been collaborating on a 3.0 update to the [Speedometer](https://webkit.org/blog/8063/speedometer-2-0-a-benchmark-for-modern-web-app-responsiveness/) benchmark. This benchmark has historically focused on TodoMVC (framework/tech-stack) implementations as one proxy for what to optimize for real-world performance. 

As part of this work, we have been rewriting and modernizing the TodoMVC implementations, such as ensuring they are using the latest versions, patterns and best-practices. Given how out of date and stale many of the "official" TodoMVC implementations are, we felt there would be high-value to the community of upstreaming this work directly to the TodoMVC repository. These implementations are going through a level of peer-review (in Speedometer) from framework experts and we are being mindful of respecting the "minimal" nature of implementations we could land here.

I wanted to express my support for this upstreaming effort and will be deferring to @flashdesignory @kara and project members like @FadySamirSadek as we begin to land this work directly in our main branch.

addyosmani:
Thanks so much for this contribution. I'm going through old issues and closing them out as we last updated the project in a major way in 2024 and want to ensure we're focused on only those issues still relevant. Thanks once again, feel free to reopen if still relevant

PR #2202, opened 2023-09-21
Add React to update folder

flashdesignory:
This version has been lifted from Speedometer and follows the WebKit styles guide.
Implementation details: React 17, functional components & hooks...

@FadySamirSadek - we'd need to see what current version(s) this would replace.

@kara


updates/react/src/todo/reducer.js
```diff
@@ -0,0 +1,62 @@
...
+    while (i--) {
+        // `| 0` is more compact and faster than `Math.floor()`.
+        id += urlAlphabet[(Math.random() * 64) | 0];
+    }
+    return id;
+}
+
+export const todoReducer = (state, action) => {
+    switch (action.type) {
+        case "ADD_ITEM":
```
FadySamirSadek, line 45: I think it might be better to use constants for action types

FadySamirSadek: I think more developers use constants to prevent typos, make it easier to refactor their code, and get auto-completion so I think it might be a good idea to use them since, WDYT ?

FadySamirSadek (changes requested):
Thank you @flashdesignory for the amazing work, I would love to get your thoughts on the comments I left. I think this should be our new official react implementation since most people now use hooks to create their react app and it should also replace the current react-hooks example but I believe we would still need an updated react app that uses class components

updates/react/src/todo/components/input.jsx
```diff
@@ -0,0 +1,49 @@
...
+    const reg = /[&<>"'/]/gi;
+    return string.replace(reg, (match) => map[match]);
+};
+
+const hasValidMin = (value, min) => {
+    return value.length >= min;
+};
+
+export function Input({ onSubmit, placeholder, label, defaultValue, onBlur }) {
+    const handleBlur = useCallback(() => {
```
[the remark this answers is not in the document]
flashdesignory: We did measure a performance gain by using it here.

updates/react/README.md
```diff
@@ -0,0 +1,40 @@
+# Speedometer 3.0: TodoMVC: React
+
+## Description
+
+This application uses React 17.0.2 to implement a todo application.
+
+-   [React](https://reactjs.org/) is a JavaScript library for creating user interfaces.
+
+## Implementation details
+
```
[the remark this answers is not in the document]
flashdesignory: I added something... let me know if that works!

updates/react/src/todo/reducer.js
flashdesignory: added constants

FadySamirSadek (approved):

PR #2219, opened 2023-12-22
Move React to Examples folder

flashdesignory:
This pr replaces the React example with the updated one.

Retired:
- React-alt
- React-hooks

@kara


FadySamirSadek:
I think we should also consider removing React+Backbone, Kotlin + React and Scala.js+React as maintaining all these different variants proved to be very challenging in the past. We can always add them back if the community requests or framework maintainers provide an updated version

FadySamirSadek (approved):

PR #2221, opened 2023-12-22
Move Angular to Examples folder

flashdesignory:
This pr replaces the Angular example with the updated one.

Retired:
- angular2
- angular2_es2015
- angularjs

@kara


FadySamirSadek:
We should also consider removing the Firebase+AngularJS example as it is very outdated as well but open to adding it back once the community or someone from the Firebase team provides an updated example

FadySamirSadek (approved):

PR #2224, opened 2023-12-28
add new badge

flashdesignory:
Add a "new" badge to all new or updated apps. 
restructured the list of examples and tweaked css.

Screenshot 2023-12-28 at 3 45 37 PM

@kara

### DIFF
--- a/index.html
+++ b/index.html
@@ -81,50 +81,103 @@
 								These are examples written in pure JavaScript.
 							</p>
 							<ul class="js-app-list-inner applist js">
-								<li class="routing">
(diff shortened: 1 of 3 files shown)


FadySamirSadek (approved):
LGTM
I think we should remove the R tab but would like to know your input on this

flashdesignory:
> LGTM I think we should remove the R tab but would like to know your input on this

I think it's not super helpful to keep it in my opinion. 
I'm fine with removing it

FadySamirSadek (approved):

PR #2257, opened 2024-08-11
React: Fix association between label and input element

mgrandrath:
The association between the `toggle-all` checkbox and its label was broken because the id value referenced in the `htmlFor` attribute was missing.


flashdesignory (approved):
thanks for the fix 🎉 . 
Could you also run `npm run build` to generate the build files?
Since it's all static output, we include the dist folder in the repository.

mgrandrath:
Done ✅

Issue #2255, opened 2024-07-01
Add sidebar to Ember example


flashdesignory:
Some newer examples did have a missing sidebar, which most of them were added back in this pr: https://github.com/tastejs/todomvc/pull/2254

Could we bring back the sidebar for the Ember example as well?

https://todomvc.com/examples/emberjs/todomvc/dist/

@NullVoxPopuli - is this something that you might be interested in adding? 😄

flashdesignory:
An example can be found here: https://todomvc.com/examples/knockoutjs/

It's part of the todomvc.com website and just gets injected in your build. 
It just shows up if you run the todomvc.com website locally, and won't affect the todomvc build. 

couple steps needed:

1. npm install todomvc-common
2. inject the css somewhere: <link rel="stylesheet" href="node_modules/todomvc-common/base.css" />
3. inject the js file somewhere: <script src="node_modules/todomvc-common/base.js"></script>
4. update this file as needed: https://github.com/tastejs/todomvc/blob/master/learn.json

To preview, you'd need to run the todomvc website from the root. 

That's pretty much it. Happy to help / review prs. 
This would be super helpful 😄

SparshMadaan:
Hi flashdesignory,this issue is still open . I'm happy to contibute if this still open .

flashdesignory:
I'll refer to @NullVoxPopuli, if offloading is desired 😄

NullVoxPopuli:
oh sorry, yes I keep forgetting about this work -- @SparshMadaan if you want to PR -- I'm more than happy to review!

SparshMadaan:
@NullVoxPopuli I'm beginner in open source and I do some little development project can you please let me know how to run this code in my local machine beacause npm start beacuse I made the changes but not able to see .

NullVoxPopuli:
Do you have errors anywhere?

SparshMadaan:
@NullVoxPopuli Thanks Buddy !

PR #2261, opened 2024-10-09
Update ember

NullVoxPopuli:
Aiming to resolve: https://github.com/tastejs/todomvc/issues/2255

How do you run the meta app?
I see engines is set to node 8...?


flashdesignory:
@NullVoxPopuli - you're right, the site needs node 8 to run with gulp.
I believe it's `npm run gulp serve` from root.

(@FadySamirSadek - in case I am forgetting something here)

NullVoxPopuli:
I can't get it to run:
```bash
❯ npm run gulp serve

> todomvc@0.1.2 gulp
> gulp serve

fs.js:43
} = primordials;
    ^

ReferenceError: primordials is not defined
    at fs.js:43:5
    at req_ (<repo>/node_modules/natives/index.js:137:5)
    at Object.req [as require] (<repo>/node_modules/natives/index.js:54:10)
    at Object.<anonymous> (<repo>/node_modules/gulp/node_modules/graceful-fs/fs.js:1:37)

```

flashdesignory:
@NullVoxPopuli  - are you using node 8.17.0?
You might have to run install again after switching, but the command is pretty much what I'm using.

NullVoxPopuli:
install never finishes

learn.json
```diff
@@ -423,17 +423,41 @@
...
 			"heading": "Community",
 			"links": [{
+				"name": "Discord",
+				"url": "https://discord.gg/emberjs"
+			}, {
+				"name": "BlueSky",
+				"url": "https://bsky.app/profile/emberjs.com"
+			}, {
+				"name": "Mastodon",
+				"url": "https://hachyderm.io/@emberjs",
```
flashdesignory, line 451: nit: trailing comma

flashdesignory (changes requested):
in your learn.json file, please update "emberjs" (on line 403) to just "ember". 
In that way it matches your index.html file (<!DOCTYPE html><html lang="en" data-framework="ember">...).

I was able to pull your branch, install and run it. Not sure why it's not working for you locally... the old node version might be a bit wonky to work with 🤷 .

NullVoxPopuli:
thanks, updated!

NullVoxPopuli:
>  Not sure why it's not working for you locally

I couldn't find what combination of npm/node to use -- I suspect that if I have the right combo of versions, I could run things

PR #2263, opened 2024-10-13
Update Mithril example from v0.2 to v2

dead-claudia:
Hand-tested all of this to work. I factored out some components for clarity, but other than that, I tried to make as few edits as I could while upgrading.

### DIFF
--- a/examples/mithril/index.html
+++ b/examples/mithril/index.html
@@ -10,8 +10,11 @@
 		<section class="todoapp"></section>
 		<footer class="info">
 			<p>Double-click to edit a todo</p>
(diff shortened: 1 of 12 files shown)


dead-claudia:
Oh, and BTW you might want to move away from Travis CI. [We got bit by being too stable to notice its shift away from freely supporting most OSS, too.](https://github.com/MithrilJS/ospec/issues/70) And no, [we're far from alone](https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works), just among the later to notice.

